PR #175, opened 2021-03-08
ERD-Generator

linusha:
The [ERD-Generator](https://github.com/beyondcode/laravel-er-diagram-generator) does not support PHP 8 as of now. I forked it and made it work (I do not know if everything works, but I was able to generate a diagram that looked correct, although we should probably tweak the config somewhat to also make it look a bit more pleasing).
I have no strong thought on whether we should merge this later on, if we do, we should keep an eye out on movement in the original repo and switch to it when they have official support for PHP 8.0 (I do not feel comfortable with preparing a PR for that, if you feel like we could try that I am on board).


tillprochaska (approved):
Nice 👍 Maybe we could add the diagram to the docs directory and possibly rename it to `er-diagram.png`? Might also be handy to have a composer script for that, something like this:

```json
"scripts": {
  "erd": "@php artisan generate:erd docs/er-diagram.png"
}
```

linusha:
Closed because of #179. Idea might be revisited if there is official PHP 8 support.

PR #187, opened 2021-03-20
150 deployment

### LINKED ISSUE #150: Deploy to Uberspace
When going public with the first visualizations in the form of sharepics (#67), there should be a single source of truth with a DB that is actually hosted and filled automatically by the scrapers.

linusha:
Close #150


tillprochaska:
We should enable [Laravel maintenance mode](https://laravel.com/docs/8.x/configuration#pre-rendering-the-maintenance-mode-view) at the beginning of the `deployment` role and disable it again on successful completion.

linusha:
What alternative do you have in mind to using `.env` files?

tillprochaska (approved):

PR #195, opened 2021-03-25
Weekly Dependabot updates

tillprochaska:
Right now, Dependabot checks for updates of PHP and Python packages every day. That’s a little annoying and tbh it’s not really important that we’re always on the latest minor version of every package. I updated the Dependabot settings to check for updates on a weekly basis from now on. (We’ll still be notified of critical updates/security patches out of schedule.)


linusha (approved):
Thank you! :heart:

PR #228, opened 2021-04-06
Increase test coverage

tillprochaska:
Prokrastination…

### DIFF
--- a/app/app/Mixins/ComponentAttributeBagMixin.php
+++ b/app/app/Mixins/ComponentAttributeBagMixin.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace App\Mixins;
+
+use Illuminate\View\ComponentAttributeBag;
+
+class ComponentAttributeBagMixin
+{
+    public function bem()
+    {
+        return function (string $base, ?string $modifiers = null): ComponentAttributeBag {
+            if (! $modifiers) {
+                return $this->class($base);
+            }
+
+            $modifiers = explode(' ', $modifiers);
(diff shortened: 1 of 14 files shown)


app/app/Mixins/ComponentAttributeBagMixin.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace App\Mixins;
+
+use Illuminate\View\ComponentAttributeBag;
+
+class ComponentAttributeBagMixin
```
linusha, line 7: Eher eine Lernfrage - wieso heißt es `bag`?

scrapers/ep_votes/api.py
```diff
@@ -47,75 +47,62 @@ def wrapper() -> Response:
     return wrapper
 
 
-@app.route("/members")
-@json_response
-def members() -> SimpleResponse:
-    term = request.args.get("term", type=int)
+def params(**params: Callable) -> Callable:
```
linusha, line 50: Do you have a link to some documentation or something at hand that broadly explains what this is about?
I see that below we use some flask decorator to define parameters for requests. I also see that this is about checking if all parameters that are needed are given, I assume that flask provides these in `params` then each time. However I do not quite get why/when this wrapper is applied. I think my question boils down to what these SimpleHandlers are, i.e. what constitures SimpleHandlers in our code below. :see_no_evil:

linusha (approved):

linusha (approved):

PR #279, opened 2021-05-23
Feature/vote view

linusha:
**To do**
- [x] Embed thumb icon as SVG sprite (instead of inline SVG)
- [x] Apply `flex-shrink: 0` to avatar and thumb
   Screenshot 2021-05-23 at 19 48 48
- [x] Fix vertical align of abstention thumb
   Screenshot 2021-05-23 at 19 44 29
- [x] Fix character encoding (the correct name seems to be **in 't Veld** (https://reneweuropegroup.eu/en/about-us/our-meps/#bloc-I)
   Screenshot 2021-05-23 at 19 37 05


linusha:
Screenshot

tillprochaska (approved):

PR #290, opened 2021-05-30
Selector assertions

### LINKED ISSUE #280: Add helpers to test for CSS selectors
See e.g. https://liamhammett.com/laravel-testing-css-selector-assertion-macros-D9o0YAQJ

tillprochaska:
This adds two PHPUnit assertion methods and the respective Pest expectations to the two relevant classes `TestView` and `TestResponse`. An instance of `TestView` ist returned e.g. when we render an isolated Blade component, e.g. like this:

```php
$view = $this->blade('<x-thumb result="adopted" />');
```

An instance of `TestResponse` is returned when we send a response in our HTTP feature test, e.g. like this:

```php
$response = $this->get('/votes/1');
```

In both cases, it would be handy to be able to assert that a particular elements exists using CSS selectors, e.g. like this:

```php
$response->assertSelector(‘.thumb—adopted.thumb--circle’);
```

Right now, we work around this using `assertSee`, e.g. like this

```php
$response->assertSee(‘thumb--adopted thumb--circle’)
```

This works, but is prone to errors. It simply asserts that the string exists somewhere in the response. As a consequence, it will break if for example the order of classes in the `class` attribute changes, the above `assertSee` matches the first element, but not the second:

```html
<span class=“thumb thumb--adopted thumb--circle”>
<span class=“thumb thumb--circle thumb--adopted”>
```

Also, using `assertSee` does not allow for more complex assertions, e.g. to test for nested elements:

```html
<li class=“list-item”>
  <span class=“thumb thumb--adopted”></span>
</li>
```

This is now possible using the newly added assertions:

```php
$response->assertSelector(‘.list-item > .thumb.thumb—adopted’)
```

Apart from that, it’s also possible to assert that a selector matches an exact amount of elements and that a selector matches elements with a given text:

```php
// Expect 3 elements with the class `.thumb--adopted`
$response->assertSelector(‘.thumb--adopted’, 3);

// Expect 2 elements with class `.list-item__subtitle` contain text `Greens/EFA`
$response->assertSelectorText(‘.list-item__subtitle’, ‘Greens/EFA’, 2);
```

As our test framework Pest basically is only a wrapper around PHPUnit, providing a more intuitive interface to write tests, I implmented these as “standard” assertions first, then added the respective Pest expectations – i.e. we can use both in our tests like this:

```php
expect($response)->toHaveSelector(‘.thumb--adopted’);
expect($response)->toHaveSelectorWithText(‘.list-item’, ‘Greens/EFA’);
```

The assertions themselves are implemented using the `css-selector-converter` package. PHP has native DOM libraries to parse XML/HTML documents using xPath expressions. The basic idea is to first convert the CSS selector to an xPath expression using `css-selector-converter`, then use the native PHP DOM libraries to query the document using the resulting xPath selector.

Closes #280.

### DIFF
--- a/app/app/Mixins/AbstractRenderedMixin.php
+++ b/app/app/Mixins/AbstractRenderedMixin.php
@@ -1,0 +1,83 @@
+<?php
+
+namespace App\Mixins;
+
+use DOMDocument;
+use DOMXPath;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Str;
+use Illuminate\Testing\Assert;
+use Symfony\Component\CssSelector\CssSelectorConverter;
+
+abstract class AbstractRenderedMixin
+{
+    abstract public function getRenderedHTML();
+
+    public function select()
+    {
+        return function (string $selector): Collection {
+            // TODO: Explain why we need to internalize/ignore errors
+            libxml_use_internal_errors(true);
+
+            $document = new DOMDocument();
+            $document->loadHTML($this->getRenderedHTML());
+
+            $converter = new CssSelectorConverter();
+            $xPathSelector = $converter->toXPath($selector);
+
+            $xPath = new DOMXPath($document);
+            $elements = $xPath->query($xPathSelector);
+
+            return collect($elements);
+        };
+    }
+
+    public function assertSelector()
+    {
+        return function (string $selector, ?int $count = null): self {
+            $elements = $this->select($selector);
+            $actualCount = $elements->count();
+
+            Assert::assertTrue(
+                $elements->isNotEmpty(),
+                "Did not find '{$selector}'."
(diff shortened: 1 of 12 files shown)


linusha:
I can follow the code in so far as that I see why it makes sense and the actual implementation of the assertions makes sense to me. I will however not pretend that I understood why it looks exactly like it does, especially that it is an abstract class that somehow seems to be able to perform some functions like  `getContent()` and has an `rendered` attribute goes over my head. 
I think this is mainly a problem with me missing some Laravel essentials. Maybe we can talk about that briefly whenever, in the end it is not urgent.

In my opinion this can be merged.

linusha (approved):

PR #403, opened 2021-08-17
Scrape summary ID based on date/week of vote

### LINKED ISSUE #401: `SummaryIDScraper`: Select ID based on date if multiple summary buttons exist

tillprochaska:
Otherwise we’d get incorrect summary IDs in case there are multiple votes with summaries for a legislative procedure.

Close #401

### DIFF
--- a/scrapers/ep_votes/scrapers.py
+++ b/scrapers/ep_votes/scrapers.py
@@ -342,7 +342,8 @@ class VoteCollectionsScraper(Scraper):
 class SummaryIDScraper(Scraper):
     BASE_URL = "https://oeil.secure.europarl.europa.eu/oeil/popups/"
 
-    def __init__(self, reference: str):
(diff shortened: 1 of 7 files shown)


scrapers/ep_votes/scrapers.py
```diff
@@ -351,12 +356,16 @@ def _url(self) -> str:
 
     def _extract_data(self) -> Optional[str]:
         rows = self._resource.select("#key_events-data .ep-table-row")
-        row = self._summary_row(rows)
+        summary_rows = [row for row in rows if self._is_summary_row(row) is True]
```
linusha, line 359: Can we name this summary_rows_in_week or something like that?
I found it to be a bit confusing at first, since the filtering for the (presumably) correct summary already takes place in `is_summary_row()`

linusha (approved):

PR #571, opened 2021-11-08
Add JSON endpoint for vote results

Alexander-Dubrawski:
Implements issue [568](https://github.com/HowTheyVote/epvotes/issues/568). 

This PR adds a new endpoint which is returning the voting results as a JSON. Moreover it adds a new button in the frontend to download the json:

Screenshot 2021-11-08 at 15 55 05

### DIFF
--- a/app/app/Http/Controllers/VotingListsController.php
+++ b/app/app/Http/Controllers/VotingListsController.php
@@ -122,6 +122,35 @@ class VotingListsController extends Controller
         return response($csv, 200, ['Content-Type' => 'text/csv']);
     }
 
+    public function json(VotingList $votingList)
+    {
+        $members = $this->members($votingList)->map(function ($member) {
+            return [
+                'member_id' => $member->id,
+                'last_name' => $member->last_name,
+                'first_name' => $member->first_name,
+                'group' => [
+                    'abbreviation' => $member->group->abbreviation,
+                    'name' => $member->group->name,
+                ],
+                'country' => [
+                    'code' => $member->country->value,
+                    'name' => $member->country->label,
+                ],
+                'position' => $member->pivot->position->label,
+                ];
+        })->toArray();
+
+        $vote = [
+            'title' => $votingList->display_title,
+            'date' => $votingList->date,
+            'result' => $votingList->result->label,
+            'members' => $members,
+        ];
+
+        return response()->json($vote);
+    }
+
     private function members(VotingList $votingList)
     {
         return $votingList
--- a/app/tests/Feature/Http/VotingLists/JsonTest.php
+++ b/app/tests/Feature/Http/VotingLists/JsonTest.php
@@ -1,0 +1,77 @@
+<?php
+
+use App\Enums\CountryEnum;
+use App\Enums\VoteResultEnum;
+use App\Group;
+use App\Member;
+use App\Vote;
+use App\VoteCollection;
+use App\VotingList;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Illuminate\Support\Carbon;
+
+uses(Tests\TestCase::class, RefreshDatabase::class);
+
+beforeEach(function () {
+    $date = new Carbon('2021-05-23');
+    $display_title = 'Budged 2020';
+
+    $greens = Group::factory([
+        'code' => 'GREENS',
+        'name' => 'Greens/European Free Alliance',
+        'abbreviation' => 'Greens/EFA',
+    ])->create();
+
+    $this->greensFor = Member::factory([
+        'first_name' => 'Jane',
+        'last_name' => 'DOE',
+        'country' => CountryEnum::NL(),
+    ])->activeAt($date, $greens)->count(1);
+
+    $this->votingList = VotingList::factory([
+        'date' => $date,
+        'vote_id' => Vote::factory([
+            'result' => VoteResultEnum::REJECTED(),
+            'vote_collection_id' => VoteCollection::factory([
+            'title' => 'Budget 2020',
+            ]),
+        ]),
+        'description' => 'Budged 2020',
+        ])
+        ->withMembers('FOR', $this->greensFor)
+        ->create();
+});
+
+it('creates a JSON with one row per member', function () {
+    $response = $this->get("/votes/{$this->votingList->id}.json")->json();
+
+    $expected = [
+        'title' => 'Budget 2020',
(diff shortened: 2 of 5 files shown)


app/tests/Feature/Http/VotingLists/JsonTest.php
```diff
@@ -0,0 +1,77 @@
...
+});
+
+it('creates a JSON with one row per member', function () {
+    $response = $this->get("/votes/{$this->votingList->id}.json");
+
+    $decoded = json_decode($response->content(), true);
+
+
+    $expected = [
+        'title' => "Children’s rights",
```
Alexander-Dubrawski, line 48: @tillprochaska we are a little bit confused here.  Where is the title `Children’s rights` coming from ? We acutely expected `Budged 2020` as a title 🙃 .

tillprochaska: You’re creating a record using a `VoteFactory`. A `VoteFactory` will by default create an associated `VoteCollection` record using the `VoteCollectionFactory` (see the [factory definition](https://github.com/HowTheyVote/epvotes/blob/main/app/database/factories/VoteFactory.php#L31)). And the `VoteCollectionFactory` in turn has a default title (in this case "Children’s rights". The whole factory setup is a little confusing, I’ll have to admit that!

So if you want to control the display title, you can configure the `VoteCollectionFactory` yourself like this:

```php
$this->votingList = VotingList::factory([
  'date' => $date,
  'vote_id' => Vote::factory([
      'result' => VoteResultEnum::REJECTED(),
      'vote_collection_id' => VoteCollection::factory([
        'title' => 'Budget 2020',
      ]),
  ]),
  'description' => 'Budged 2020',
])->withMembers('FOR', $this->greensFor)->create();
```

app/app/Http/Controllers/VotingListsController.php
```diff
@@ -122,6 +122,40 @@ public function csv(VotingList $votingList)
...
+
+        $voting= [
+            'title' => $votingList->display_title,
+            'date' => $votingList->date,
+            'result' => $votingList->result->label,
+            'members' => $members,
+        ];
+
+
+        ob_start();
```
tillprochaska, line 152:
```suggestion
```

This is not necessary in this case. `ob_start` caches `stdout` and is used in the `csv` endpoint as a workaround, as the `SimpleExcelWriter` only provides methods to write to a file. However, such a workaround is not necessary in this case.

app/app/Http/Controllers/VotingListsController.php
```diff
@@ -122,6 +122,40 @@ public function csv(VotingList $votingList)
...
+
+        $json = json_encode($voting);
+
+        return response($json, 200, ['Content-Type' => 'application/json']);
```
tillprochaska, line 156: This works 100% as expected. Laravel does however provide a few little convenience helpers to return JSON responses which are a little more idiomatic. The helper will handle setting the correct content type headers and status code for you.

Relevant docs: https://laravel.com/docs/8.x/responses#json-responses

```suggestion
        return response()->json($voting);
```

tillprochaska: And another small thing: The terminology is a little confusing and I’ll add a small glossary section to the docs:

* `Voting`: How a single member voted, e.g. "David Sasolli voted FOR in vote 123"
* `Vote`: A single vote, containing information like the date and result (e.g. `ADOPTED`/`REJECTED` for a single vote. Belongs to a single `VotingList`.
* `VotingList`: Has many `Votings` and belongs to a single `Vote`.
* `Vote`s and `VotingList`s basically represent the same "real-life object", but are stored in separate database tables, because we source them from different parliamentary records that become available at different times.
* `VoteCollection`: A collection of many related `Vote`s.

tillprochaska: So in this case the variable `$voting` actually is a `$vote` or `$votingList`.

app/tests/Feature/Http/VotingLists/JsonTest.php
```diff
@@ -0,0 +1,77 @@
...
+
+
+    $expected = [
+        'title' => 'Children’s rights',
+        'date' => '2021-05-23T00:00:00.000000Z',
+        'result' => 'rejected',
+        'members' => [
+            [
+                // member_id is autogenerated inside the database, because it is the first entry.
+                'member_id' => 1,
```
tillprochaska, line 54: Yes, in theory, this *should* be the case, as the test database *should* be reset before each test run. However, in the past, this led to problems. So I’d not use a constant ID here, but instead do something like this:

```suggestion
                // member_id is autogenerated inside the database, because it is the first entry.
                'member_id' => $this->votingList->members()->first()->id,
```

app/tests/Feature/Http/VotingLists/JsonTest.php
```diff
@@ -0,0 +1,77 @@
...
+        'description' => 'Budged 2020',
+        ])
+        ->withMembers('FOR', $this->greensFor)
+        ->create();
+});
+
+it('creates a JSON with one row per member', function () {
+    $response = $this->get("/votes/{$this->votingList->id}.json");
+
+    $decoded = json_decode($response->content(), true);
```
tillprochaska, line 44: There’s a Laravel helper method for this as well. You can simply do the following:

```suggestion
    $response = $this->get("/votes/{$this->votingList->id}.json")->json();
```

tillprochaska:
Thanks a lot, this is super helpful (the feature itself, as well as the feedback on how the readme/docs could be improved)!

I added a few comments/suggestions. Most of them Laravel-specific! :)

app/app/Http/Controllers/VotingListsController.php
linusha: @tillprochaska those comments actually do exist, but we could debate migrating them into the readme or something like that. We explicitly decided against that earlier. @Alexander-Dubrawski you can find (hopefully helpful) descriptions for all models in the corresponding php files in the model folder. :)

Alexander-Dubrawski: It is always helpful to have a description in the corresponding folders and files :). I also would find a diagram like for example:

class-diagram

helpful. This does not need to be directly inside the readme, but maybe in a wiki? 🙃

tillprochaska (approved):

PR #647, opened 2022-04-02
Fix `composer install` in CI/CD

tillprochaska:
Composer recently introduced a change that will prompt you to confirm to install any plugins that come with packages. This is the case for pest, so our CI runs failed after reaching a timeout, because composer waited for user input.

The solution is to add the pest plugin to our `composer.json`, and in general (and to prevent similar issues in the future) to run composer commands non-interactively in CI/CD.


linusha (approved):
deployment

Issue #466, opened 2021-08-24
Set up browser testing


tillprochaska:
See: https://github.com/HowTheyVote/epvotes/commit/3440402dd81c038cf52c139ff394507e578ee846

PR #707, opened 2022-08-05
Members votes search

tillprochaska:
Newer versions of Meilisearch support nested objects. This allows us to reuse our existing search UI (which fetches votes directly from the search index). This has two advantages:

1) The votes displayed on member profiles are often out of date, because we cache them statically. When dynamically rendering votes, the most recent votes will always be “fresh”, as they are fetched from the search index.

2) We can search for a certain keyword or topic and see at a glance how an MEP voted in multiple votes matching the search term.

With nested index objects, we can index a vote like this (where 1234 and 5678 are the IDs of members):

```
{
  "title": "Lorem Ipsum",
  "result": "ADOPTED",
  …
  "members": {
    "1234": "FOR",
    "5678": "AGAINST",
    …
  }
}
```

We can then retrieve only the voting for a single member from the search index:

```
{
  "q": "Lorem",
  ...
  "attributesToRetrieve": [
    "title",
    "members.1234",
    ...
  ]
}
```

As this PR includes an upgrade of Meilisearch, we need to run the respective Ansible playbook to build the new Meilisearch version and reimport the `voting_lists` index afterwards.

I’ve also included two changes in this PR that aren’t directly related:

* I’ve upgrade the Docker base image to use the latest version of Alpine and PHP and I’ve updated composer dependencies that weren’t compatible with PHP 8.1 in the currently locked version.
* I’ve upgraded the code style presets we use, because the original package isn’t maintained anymore which caused our CI to fail. There’s a maintained fork by a new maintainer.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
     - name: Setup PHP
       uses: shivammathur/setup-php@v2
       with:
-        php-version: '8'
+        php-version: '8.1'
         coverage: pcov
 
     - uses: actions/checkout@v2
--- a/app/app/VotingList.php
+++ b/app/app/VotingList.php
@@ -146,16 +146,16 @@ class VotingList extends Model
         }
 
         return __('voting-lists.share-picture.alt-text', [
-           'title' => $this->display_title,
-           'date' => $this->date->formatLocalized('%b%e, %Y'),
-           'for' => $this->stats['by_position']['FOR'],
-           'forpercent' => round(($this->stats['by_position']['FOR'] / $this->stats['voted'] * 100)),
-           'against' => $this->stats['by_position']['AGAINST'],
-           'againstpercent' => round(($this->stats['by_position']['AGAINST'] / $this->stats['voted'] * 100)),
-           'abstention' => $this->stats['by_position']['ABSTENTION'],
-           'abstentionpercent' => round(($this->stats['by_position']['ABSTENTION'] / $this->stats['voted'] * 100)),
-           'voted' => $this->stats['voted'],
-           'novote' => $this->stats['by_position']['NOVOTE'],
+            'title' => $this->display_title,
+            'date' => $this->date->formatLocalized('%b%e, %Y'),
+            'for' => $this->stats['by_position']['FOR'],
+            'forpercent' => round(($this->stats['by_position']['FOR'] / $this->stats['voted'] * 100)),
+            'against' => $this->stats['by_position']['AGAINST'],
+            'againstpercent' => round(($this->stats['by_position']['AGAINST'] / $this->stats['voted'] * 100)),
+            'abstention' => $this->stats['by_position']['ABSTENTION'],
+            'abstentionpercent' => round(($this->stats['by_position']['ABSTENTION'] / $this->stats['voted'] * 100)),
+            'voted' => $this->stats['voted'],
+            'novote' => $this->stats['by_position']['NOVOTE'],
         ]);
     }
 
--- a/app/app/VotingList.php
+++ b/app/app/VotingList.php
@@ -73,6 +73,9 @@ class VotingList extends Model
             'summary' => $this->vote->voteCollection->summary?->text,
             'session_id' => $this->vote->session?->id,
             'session_display_title' => $this->vote->session?->display_title,
+            'members' => $this->members->mapWithKeys(function ($member) {
+                return [$member->id => $member->pivot->position->label];
+            }),
         ];
     }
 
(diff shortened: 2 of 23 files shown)


app/app/VotingList.php
```diff
@@ -73,6 +73,9 @@ public function toSearchableArray()
             'summary' => $this->vote->voteCollection->summary?->text,
             'session_id' => $this->vote->session?->id,
             'session_display_title' => $this->vote->session?->display_title,
+            'members' => $this->members->mapWithKeys(function ($member) {
```
tillprochaska, line 76: We should add `members` to `makeAllSearchableUsing` in order to eagerly load the relation.

tillprochaska: I tried adding `members` to the list of relations to load eagerly, but for some reason, it didn’t import/index any voting lists at all. It did not error though. I’d probably need to take a look at the generated SQL, as I do not really have a clue why that’s happening…

.github/workflows/ci.yml
```diff
@@ -90,7 +90,7 @@ jobs:
     - name: Setup PHP
       uses: shivammathur/setup-php@v2
       with:
-        php-version: '8'
+        php-version: '8.1'
```
linusha, line 93: I went through the Changelog for 8.1, there is now native Enumeration in PHP.
[This discussion](https://github.com/spatie/enum/issues/90) might also be relevant.
I will look into this, but if you already thought about this and strongly want to keep the spatie package, feel free to intervene. In any case, only tangential related to this PR...

app/resources/js/components/search-results-page.js
```diff
@@ -3,7 +3,6 @@ const ATTRIBUTES = [
   'id',
   'display_title',
   'date',
-  'result',
```
linusha, line 6: Probably a stupid question - why is this gone?

app/resources/js/components/search-results-page.js
```diff
@@ -108,6 +108,11 @@ export default (options = {}) => ({
     this.loading = true;
     this.abortController = new AbortController();
 
+    const additionalAttributes =
+      this.memberId !== null ? [`members.${this.memberId}`] : ['result'];
+
+    console.log(this.memberId);
```
linusha, line 114: Is this kept on purpose?

linusha (commented):
Tests dafür zu haben (zumindest irgendwie "wird mit Stimmverhalten des MEP angezeigt" wäre schon cool eigentlich. Das ist quasi durch #466 abgedeckt, oder?

app/resources/js/components/search-results-page.js
linusha: Or rather - was this unused before?

.github/workflows/ci.yml
tillprochaska: I’m definitely in favor of using native enums over a library, but so far, I’ve been too lazy to look into what’s necessary to migrate… :)

app/resources/js/components/search-results-page.js
tillprochaska: Good catch!

app/resources/js/components/search-results-page.js
tillprochaska: It’s not gone completely, but it’s now only included when fetching data for `/votes`. When feting data for `/members/:id`, we retrieve the `members.${id}` attribute instead.

https://github.com/HowTheyVote/epvotes/pull/707/files#diff-e6903bf1e45a09bd1417c6051799566378f9a71210359b0673f6578263e8a194R111-R112

tillprochaska:
> Tests dafür zu haben (zumindest irgendwie "wird mit Stimmverhalten des MEP angezeigt") wäre schon cool eigentlich. Das ist quasi durch https://github.com/HowTheyVote/epvotes/issues/466 abgedeckt, oder?

Ja, genau

app/resources/js/components/search-results-page.js
tillprochaska: The code is somewhat confusing though and could probably be simplified to something like this:

```js
const attributesToRetrieve = [...ATTRIBUTES, `members.${this.memberId}` || 'result'];
```

or

```js
const attributesToRetrieve = this.memberId
  ? [...ATTRIBUTES, `members.${this.memberId}`]
  : [...ATTRIBUTES, 'result']
```

linusha (approved):

app/resources/js/components/search-results-page.js
linusha: Got it, thanks. 
Personally, I did not find the other snippets to be easier to read than the code as is.

PR #709, opened 2022-08-08
Update to python3.10 and resolve dependency problems / Stop using werkzeug for prod

linusha:
The change in the `Dockerfile` resolves the problem that I mentioned where `poetry` will insist on using outdated envs on systems where the container existed previously. On a "fresh" clone/machine the dependency issues we talked about where non-existent.

I already tried starting the `gunicorn` server instead of the `flask` one locally and made sure that I can run scrapers.
I propose that I deploy this branch once after you gave :heavy_check_mark: and before merging, to make sure that everything works remote as expected.

By the way, if you test this locally, rebuilding the container, even without any caches, was not enough for me, I had to also go into the container and run another `poetry install`.

### DIFF
--- a/scrapers/Dockerfile
+++ b/scrapers/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.8-alpine3.12
+FROM python:3.10-alpine3.16
 
 RUN apk --update add \
     build-base \
--- a/scrapers/Dockerfile
+++ b/scrapers/Dockerfile
@@ -19,7 +19,7 @@ WORKDIR /app
 
 COPY pyproject.toml pyproject.toml
 COPY poetry.lock poetry.lock
-RUN poetry env use python
+RUN poetry env use python3.10
 RUN poetry install
 
 COPY . .
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,9 +13,9 @@ jobs:
 
     steps:
     - name: Set up Python
-      uses: actions/setup-python@v2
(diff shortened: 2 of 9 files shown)


scrapers/Dockerfile
```diff
@@ -19,7 +19,7 @@ WORKDIR /app
 
 COPY pyproject.toml pyproject.toml
 COPY poetry.lock poetry.lock
-RUN poetry env use python
+RUN poetry env use python3.10
```
tillprochaska, line 22: @linusha Any idea as to why this is required? Shouldn’t `python` link to `python3.10` in the container, i.e. `python -v` == `python3.10 -v`?

linusha: Not really unfortunately. As I wrote above, this was necessary in order to get poetry to stop using the older python3.9 venv inside of the container. But I am a) not sure why it is necessary, as I would have had the same expectation as you and b) I am also confused as to why there still were older poetry venvs and stuff after deleting my scrapers container and rebuilding it without any caches.

tillprochaska: Maybe because we persist the venvs in a volume. Still, while building the image, `poetry install` should have been run… Will try it out later and see if this is an issue for me as well!

linusha:
@tillprochaska I pushed some minor, additional fixes. But I could deploy this branch and scrape sessions without causing errors in sentry or otherwise. So I think we are good to go.

tillprochaska (approved):

PR #726, opened 2022-08-11
Agenda links

linusha:


app/tests/Unit/Components/SessionDisplayTest.php
```diff
@@ -42,6 +42,16 @@
...
 });
 
+it('displays link to the agenda of the session', function () {
+    $view = $this->blade('<x-session-display :currentSession="$currentSession" :lastSession="$lastSession" :nextSession="$nextSession" />', [
+        'currentSession' => $this->currentSession,
+        'lastSession' => $this->lastSession,
+        'nextSession' => $this->nextSession,
+    ]);
+
+    expect($view)->toSee('https://www.europarl.europa.eu/doceo/document/OJ-9-2021-08-01-SYN_EN.html');
```
tillprochaska, line 52: Maybe we can be a little more specific?

```suggestion
    expect($view)->toHaveSelector('a[href="https://www.europarl.europa.eu/doceo/document/OJ-9-2021-08-01-SYN_EN.html"]', text: 'agenda');
```

app/resources/views/components/session-display.blade.php
```diff
@@ -5,13 +5,16 @@
...
- @if ($currentSession)
+@if ($currentSession)
 
     <div {{ $attributes->bem('session-display') }}>
         {{ __('components.session-display.current', [
             'location' => $currentSession->location->label,
         ]) }}
 
+        {!! __('components.session-display.agenda', [
+            'agenda' => $currentSession->agendaLink]) !!}</br>
```
tillprochaska, line 16: While the naming convention is to use camel case for method names, attributes are usually referred to using snake case.

I think Laravel is smart enough to call the `getAgendaLinkAttribute` accessor for the attribute `agenda_link`, but I’m not 100% sure.

```suggestion
            'agenda' => $currentSession->agenda_link]) !!}</br>
```

app/resources/lang/en/components.php
```diff
@@ -45,5 +45,6 @@
         'current' => 'Parliament is meeting in :location this week.',
         'last' => 'The last plenary session was held <strong>from :start to :end</strong> in :location.',
         'next' => 'The next plenary session will be held in :location <strong>from :start to :end</strong>.',
+        'agenda' => 'Take a look at this session\'s votes in its <a href=":agenda" target="_blank" rel="noopener noreferrer">agenda</a>.',
```
tillprochaska, line 48: I have a suggestion for slightly easier wording:

> Take a look at [the agenda]() for this session.

Or something like this, although we would have to include two variants for upcoming/past sessions:

> Take a look at [the agenda]() to find out what MEPs have voted on.
> Take a look at [the agenda]() to find out what MEPs will be voting on.

app/app/Session.php
```diff
@@ -72,4 +72,9 @@ public function votes()
     {
         return $this->hasMany(Vote::class);
     }
+
+    public function getAgendaLinkAttribute()
```
tillprochaska, line 76: This is nitpicking, but in other places we’re using the term "url" over "link", so how about this?

```suggestion
    public function getAgendaUrlAttribute()
```

app/resources/views/components/session-display.blade.php
linusha: My test is still green, so I assume Laravel is indeed smart enough. Nice, I did not know this!

tillprochaska (approved):

PR #727, opened 2022-08-11
Track database dumps with git LFS

linusha:


ansible/roles/deployment/tasks/main.yml
```diff
@@ -5,6 +5,10 @@
     path: "{{ epvotes_install_dir }}/app/artisan"
   register: artisan_command
 
+- name: Setup Git LFS
+  shell:
+    cmd: git lfs install
+
```
tillprochaska, line 11: Can we somehow not install LFS and simply exclude LFS files when cloning on Uberspace? Like we don’t need the database dumps there…

linusha: Ah, you are right - the dump gets committed from inside the runner!
From what I understand is that we can actually just delete this step again and everything will work, as long as we do not touch the contents of the dumps.
But this makes me think - do we have to install git lfs inside of the action runner?

tillprochaska: Yeah, we’d have to set the `lfs` option for the `checkout` action in the `dump` workflow:

https://github.com/actions/checkout

PR #741, opened 2022-09-04
Scrape pictures for members without picture in scrape:all

linusha:
The missing pictures in production are already fixed :)

image

### DIFF
--- a/app/app/Member.php
+++ b/app/app/Member.php
@@ -123,6 +123,11 @@ class Member extends Model
         return Storage::disk('public')->url("members/{$this->id}.jpg");
     }
 
+    public function hasProfilePicture(): bool
+    {
+        return Storage::disk('public')->exists("members/{$this->id}.jpg");
+    }
+
     public function getLinksAttribute(): Collection
     {
         return collect([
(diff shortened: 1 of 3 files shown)


app/app/Member.php
```diff
@@ -123,6 +123,11 @@ public function getProfilePictureUrlAttribute(): string
         return Storage::disk('public')->url("members/{$this->id}.jpg");
     }
 
+    public function getProfilePictureExistsAttribute(): bool
```
tillprochaska, line 126: Just an idea, but couldn’t we cover both use cases with a single method?


```php
public function getProfilePictureUrlAttribute(): ?string
{
  $path = "members/{$this->id}.jpg";
  $disk = Storage::disk('public');

  if (!$disk->exists($path)) {
    return null;
  }

  return $disk->url($path);
}
```

In the Artisan command, you could still filter the members in the exact same way.

tillprochaska (approved):

app/app/Member.php
linusha: To be honest I would prefer to keep it this way. I came to not like "overloading" methods with regards to their return type.
If you would like that really much we can change it, as I absolutely see arguments in favor of it and my skepticism might be formed mainly by working a lot without type hints.
But otherwise, I would proceed :see_no_evil:

tillprochaska: Totally fine as well – can we name it following the convention we use for other similar methods in that case, for example `hasProfilePicture()`?

https://github.com/HowTheyVote/epvotes/blob/main/app/app/Vote.php#L141

linusha: @tillprochaska Done, if everything works for you, feel free to just merge this :) Thanks!

PR #749, opened 2022-09-06
Remove dumps, related action and LFS references, as we exhaust quotas

linusha:


tillprochaska (approved):
Can we add back the dev dump as a regular, non-LFS file? It should be rather small and doesn't change that often

linusha:
> Can we add back the dev dump as a regular, non-LFS file? It should be rather small and doesn't change that often

Yes, most definitely! However, I was unable to get the actual file back from LFS (I think because we exhausted the quotas or something). I will retrieve the dev dump from an older commit and add it back, did not yet do it due to time constraints and have that on my list for the PR that also introduces the stuff for the new dump archiving :slightly_smiling_face:

PR #756, opened 2022-09-20
Fixup missing 0 to construct isoformat date string

linusha:
The 0 went missing due to typeconversions, i.e. even if we supplied the initial values with leading 0, when converting them to ints, they went amiss.

Fixes https://sentry.io/share/issue/a93a8620ca21413d9d5b5f7027b3a399/.

### DIFF
--- a/scrapers/ep_votes/scrapers.py
+++ b/scrapers/ep_votes/scrapers.py
@@ -611,7 +611,13 @@ class ParliamentSessionsScraper(Scraper):
         )
 
     def _parse_date(self, day: str) -> date:
-        return date.fromisoformat(str(self.year) + "-" + str(self.month) + "-" + day)
+        return date.fromisoformat(
+            str(self.year).rjust(2, "0")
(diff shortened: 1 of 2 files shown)


scrapers/ep_votes/scrapers.py
```diff
@@ -611,7 +611,13 @@ def _session(self, session: dict) -> Session:
         )
 
     def _parse_date(self, day: str) -> date:
-        return date.fromisoformat(str(self.year) + "-" + str(self.month) + "-" + day)
+        return date.fromisoformat(
+            str(self.year).rjust(2, "0")
+            + "-"
+            + str(self.month).rjust(2, "0")
+            + "-"
+            + day
```
tillprochaska, line 619: Should we maybe add leading zeros here as well, just in case that the source we crawl changes its format?

```suggestion
            + day.rjust(2, "0")
```

tillprochaska (approved):

PR #758, opened 2022-09-20
Add switch to manually exempt Sessions from voting list scraping

linusha:
This fixes a crash that occurs when we try to scrape the voting lists for a Session that has no voting lists in scrape:all. In the future, we can manually mark such Sessions. I opted for manual marking as the implications of exempting a Session are quite severe and this case is very rare.


linusha:
@tillprochaska whoever deploys this, afterwards we need to set this attribute to true for the Session with id==19`. Then, the next `scrape:all` run should not throw errors that are related to voting lists that cannot be found. :)

tillprochaska (approved):

Issue #751, opened 2022-09-10
Newer Members of Parliament are not Sorted Alphabetically When Listing all Members


linusha:
Go on a vote site and expand the list of displayed members.
All (I just assume that our theory wrt the pictures is correct) newer members of parliament are at the bottom, despite that clearly not being the alphabetical order.

![image](https://user-images.githubusercontent.com/14252419/189479707-84d0ea18-4e13-4587-b826-cfe2acef53ce.png)

PR #767, opened 2022-09-29
Sort members of `VotingList` alphabetically

linusha:
This should fix the issue everywhere where we use `->members` of a `VotingList`. Fixes e.g., #751.

The testing of this is kind of meh.
We could either make a real VotingListController Test which would make sense imo, or beef the other two or three tests where this is used up as well. Especially the votinglist/show test. However, to do this well I would have needed to create `expectSelectorTextInOrder`.


tillprochaska:
Maybe it would make sense to use a [global scope](https://laravel.com/docs/9.x/eloquent#global-scopes) to ensure that members are always ordered alphabetically (if not explicitly ordered otherwise)?

linusha:
> Maybe it would make sense to use a [global scope](https://laravel.com/docs/9.x/eloquent#global-scopes) to ensure that members are always ordered alphabetically (if not explicitly ordered otherwise)?

Nice, I forgot about those :slightly_smiling_face:

app/app/Http/Controllers/VotingListsController.php
```diff
@@ -155,7 +155,7 @@ public function json(VotingList $votingList)
         return response()->json($vote);
     }
 
-    private function members(VotingList $votingList)
+private function members(VotingList $votingList)
```
tillprochaska, line 158: Unintended change?

tillprochaska (approved):
Nice 👍 There’s a weird formatting change in one file, I guess that’s unintended, apart from that feel free to merge :)

PR #805, opened 2022-11-05
Misc meilisearch fixes

tillprochaska:


tillprochaska:
>  I would like to have an easy way of setting the same settings locally.

You can run Ansible playbooks against localhost, so the cleanest option imo would be to extract the tasks that configure the index settings into a separate playbook -- so that there’s one playbook to install Meilisearch and another to configure the index. Then we can easily run the second playbook to configure the index on a local Meilisearch instance.

tillprochaska:
In theory yes, at least given that you can build the new version in the same way you can build the old version…

https://github.com/HowTheyVote/meilisearch-centos7/pull/1

tillprochaska:
> If you hold back merging this, I can set that up and append it to this PR.

@linusha I think we can/should still do this, but I think I won’t be able to do it soon, so maybe we can merge this before that change. Apart from that, this should be ready for review.

linusha (approved):

PR #808, opened 2022-11-05
Show placeholder image when we do not have a thumbnail/profile picture

linusha:


app/app/Member.php
```diff
@@ -113,14 +113,18 @@ public function scopeWithGroupMembershipAt(Builder $query, \DateTime $date)
         });
     }
 
-    public function getThumbnailUrlAttribute(): string
+    public function getThumbnailUrlAttribute(): ?wstring
```
tillprochaska, line 116:
```suggestion
    public function getThumbnailUrlAttribute(): ?string
```

tillprochaska (approved):

app/app/Member.php
linusha: Upsi.

PR #883, opened 2023-05-02
FIx `ParliamentSessionsScraper` for Sessions that span multiple months

### LINKED ISSUE #882: The session on the starting page is not correct and non-sensical
This is what we currently display:

Screenshot from 2023-05-02 14-08-57

This is the calendar:
image

linusha:
We previously assumed, that a session never spans multiple years or months. This assumption does not hold, and the first such session that we  encountered lead to a faulty session entry that in turn got displayed on the landing-page in a nonsensical way. This removes the assumption of a session never spanning multiple months. For years, we still keep our assumption.

Closes #882.


tillprochaska (approved):
🎉🙌

PR #903, opened 2024-05-24
v2

tillprochaska:
In the run-up of the European elections in June 2024, we have completely overhauled HowTheyVote.eu. We have significantly improved data quality and completeness, and vote results will generally be available on HowTheyVote.eu much faster – usually within 30-60 minutes after a vote took place.

This initially started out as a separate experiment that slowly turned into a mostly feature-complete replacement of the existing HowTheyVote.eu website. Unfortunately, we weren’t able to completely transfer the commit history.


docs/data-sources.md
tillprochaska, line 1: This needs to be updated

linusha (approved):
The other stuff I would've liked to include is already mentioned on /about.

I actually think that including the funding information is strictly necessary. Even if not, arguments by authority are great.

PR #965, opened 2024-06-22
Add CSV export

tillprochaska:
This implements a `htv export` command that exports the database to CSV tables. All relevant database tables are exported as (mostly) normalized CSV tables (e.g. `members.csv`, `votes.csv`, `member_votes.csv`).

The command also creates a metadata file that describes the table schema following the [CSV on the Web standard](https://www.w3.org/TR/tabular-data-primer/) and a markdown Readme that describes the table schema ([example](https://gist.github.com/tillprochaska/8aee64134d6ede85bc6240e69c4d9cda)). The implementation is similar to the implementation of the OpenAPI specs: Rows of each table are modeled using a Python TypedDict, for example:

```python
class MemberRow(TypedDict):
    """Each row represents a Member of the European Parliament (MEP)."""

    id: int
    """Member ID as used by the [MEP Directory](https://www.europarl.europa.eu/meps/en/home)."""

    first_name: str
    """First name"""

    last_name: str
    """Last name"""

    country_code: str
    """3-letter ISO-3166-1 code"""

    # ...
```

The type annotations and docstrings are used to auto-generate metadata and documentation.

### Publishing

The `htv export` command creates multiple CSV and metadata files and stores them on the local file system. I’ve configured Caddy to serve these files at `https://howtheyvote.eu/export`.

My idea was to then create a separate public GitHub repository with a scheduled Actions workflow that downloads the files and commits them to the repository or publishes them as a release.

This way, we don’t have to (manually) handle credentials for downloading the export or committing it to the repository.

### Data coverage

The export is currently still missing some data:

* Information about relevant geographic areas and EuroVoc concepts for votes. Should be easy enough to add that in a second step, but I wanted to focus on the most frequently requested data first to keep the scope of the PR small.
* Press releases
* Information about sources/fragments. Not sure there’s a use case for that, I’d suggest not to implement that for now.

### Todos

- [x] Double-check documentation for tables/columns. I think a little more detail would be helpful in some cases, for example for `is_main` etc.
- [x] Generate export in regular interval (e.g. weekly) and make it publicly available. We could then easily set up a GitHub Action to pull and commit the export in a public repository (similar to what we used to do before).
- [x] Implement Prometheus metric for successful export generations (so we can set up an alert if it fails).


backend/howtheyvote/export/__init__.py
```diff
@@ -0,0 +1,357 @@
...
+    """Each row represents how an MEP voted in a roll-call vote."""
+
+    vote_id: int
+    """Vote ID"""
+
+    member_id: int
+    """Member ID"""
+
+    position: str
+    """Vote position"""
```
linusha, line 154:
```suggestion
    """FOR, AGAINST, and ABSTENTION can be actively chosen by the MEP in Plenary. Is DID_NOT_VOTE if the MEP was not present in plenary at the time of the vote."""
```

backend/howtheyvote/export/__init__.py
```diff
@@ -0,0 +1,357 @@
...
+    is empty, this falls back to the procedure title."""
+
+    reference: str | None
+    """Reference to a plenary document such as a report or a resolution"""
+
+    description: str | None
+    """Description of the vote as published in the roll-call vote results"""
+
+    is_main: bool
+    """Whether this vote is a main vote."""
```
linusha, line 130:
```suggestion
    """Whether this vote is a main vote. These take place after all amendments for a matter have been voted on. They usually result in a final position of parliament on a matter."""
```

linusha:
1. Thanks for implementing this! 😊
2. I have extended the explanations for is_main and position a tiny bit. I think we could try with these as a first step and still expand them when more questions are still coming.
3. I will setup the data repository we already have once we merged this. I like your idea with the download, though. That's smart.

linusha (approved):

PR #1009, opened 2024-08-06
Add term column for votes

tillprochaska:
We know the term for every vote we scrape, but we currently do not store it in the `votes` table.

Being able to filter votes by term is useful, for example in order to create exports separately for every term (so they don’t become huge), to search within votes of the current term, …

While it’s possible to derive the term from the date, simply storing it as a separate column is much easier :)

- [x] Add separate term column
- [x] Add CLI command to back-fill term for existing votes
- [ ] Add term column to export?


linusha:
>  Add term column to export?

What about splitting the export from the get-go? Meaning having one for Term 9 and one for Term 10, currently...

PR #1016, opened 2024-08-16
Implement search using Xapian

tillprochaska:
This replaces Meilisearch with Xapian, a Lucene-style search library. The main advantages are:

* Proper BM25 search with flexible ranking/scoring customization (e.g. this allows us to rank recent or featured votes higher).
* Xapian is a library, not a search server. We don’t have to run a separate search server.

- [x] Implement filtering (see failing test `test_search_query_handle_filters`)
- [x] Implement document/procedure reference recognition (see failing test `test_votes_api_search_references`)
- [x] Implement default sort order (see failing test `test_votes_api_search`)
- [x] Add type declarations

Test locally:

```
docker compose build backend
docker compose run --rm backend poetry install
docker compose run --rm backend htv system upgrade
docker compose run --rm backend htv aggregate votes
docker compose up -d backend
```

Deployment:

* Probably the Meilisearch container and storage directory need to be removed manually. (Although it also isn’t really a problem if that isn’t done directly after deployment.)
* After deployment, the initial search index has to be built manually using `htv aggregate votes`.


backend/howtheyvote/api/query.py
```diff
@@ -168,80 +188,92 @@ def where(self, expression: ColumnElement[Any]) -> Self:
...
+    See: https://getting-started-with-xapian.readthedocs.io/en/latest/advanced/postingsource.html
+    """
+
+    def set_max_diff(self, max_diff: float) -> None:
+        self.max_diff = max_diff
+
+    def set_origin(self, origin: float) -> None:
+        self.origin = origin
+
+    def get_weight(self) -> float:
```
tillprochaska, line 212: Not sure how efficient this is, as I assume there’s a performance overhead due to Xapian having to call a Python method for every matching document during ranking. But I think at our scale this is pretty much irrelevant.

.github/workflows/backend.yml
tillprochaska, line 1: Note to self: Changes in this workflow need to be applied to the `scheduled` workflow as well

tillprochaska, line 1: Note to self: Changes in this workflow need to be applied to the scheduled workflow as well

tillprochaska:
@linusha I fixed the issue we were seeing (see details in https://github.com/HowTheyVote/howtheyvote/commit/0a25e8600f5aa6fc155db9b63c1794c1754c89ee). It should now work on your machine, too.

backend/howtheyvote/api/query.py
```diff
@@ -254,3 +298,96 @@ def query(self, query: str | None = None) -> Self:
...
+    def _xapian_bm25_weight(self) -> Weight:
+        # This parameter controls document length normalization. As we currently index mostly
+        # short vote titles and a few keywords, we set b=0 to disable normalization. We do
+        # not want a vote to score higher just because its title is shorter. If that changes
+        # (and we start indexing longer documents), we should reconsider this.
+        b = 0  # Xapian default: 0.5
+
+        # The following BM25 parameters are Xapian’s defaults.
+        # See: https://xapian.org/docs/apidoc/html/classXapian_1_1BM25Weight.html
+        k1 = 1
```
tillprochaska, line 388: Maybe adjust this to give less weight to votes just because they contain a search term multiple times. For example, when searching for "Ukraine", a vote that contains Ukraine in the title and as a geographic area will always be ranked higher than a vote that contains it only in the title, even if the latter is more recent. Not sure this is justified.

.github/workflows/backend.yml
tillprochaska: Done

backend/howtheyvote/api/query.py
linusha: This sounds reasonable, maybe we should start with `k1 = 0 and go from there, as your argument makes sense?

Issue #1022, opened 2024-08-27
Vote results chart accessibility issues


tillprochaska:
- [x] Increase contrast of striped background pattern to make the different bars easier to differentiate for low-vision and color blind users.
- [x] Increase height of slim bar charts in the "Countries" and "Political groups" tabs.
- [x] Increase gap between the different bars.

Issue #1021, opened 2024-08-27
Vote positions not announced correctly by all screen readers


tillprochaska:
The `span` element should have an extra `aria-label="against"` attribute as some screen readers ignore empty elements (even though the element has a `title` attribute).

```html
<span class="thumb thumb--against thumb--circle " title="against"><svg aria-hidden="true"><use href="/static/icons.svg#thumb"></use></svg></span>
```

PR #1024, opened 2024-08-27
Enhancement/accessibility

### LINKED ISSUE #1023: Tab buttons not keyboard accessible
Tab buttons are currently not accessible via the tab key (only via arrow keys). Remove `tabindex="-1"`.

linusha:
Fixes #1023, #1021, #1022.

---

Before:

image

After:
image
---

@tillprochaska contrast comes down to around 1.5:1-1.6:1 now. I opted for the same "level of whiteness" for both green and red, and then just made it high enough so that the contrast works for both colors. Having different "whiteness" levels looked weird to me, but I imagine you might have thoughts :upside_down_face:.

Also, I am unsure with the gaps - first question is if those even are the gaps that were meant? Or have I completely misunderstood the problem? If those are the correct gaps - the result looks weird to me, but I think that will always be the case once we need to add space between them. Might just be a question of familiarization on my end.

### DIFF
--- a/frontend/src/components/VoteResultChart.css
+++ b/frontend/src/components/VoteResultChart.css
@@ -1,21 +1,15 @@
 .vote-result-chart__bars {
   display: flex;
+  gap: 2px;
   min-height: calc(2 * var(--space-xs));
   border-radius: var(--border-radius-sm);
   overflow: hidden;
-  background-color: var(--gray-light);
-  box-shadow: inset 0 0 2px rgba(0, 0, 0, 0.25);
   margin-bottom: var(--space-xxs);
-}
-
-.vote-result-chart__bars > * + * {
-  margin-left: 1px;
 }
 
 .vote-result-chart__bar {
   width: calc(var(--ratio) * 100%);
   padding: var(--space-xs) 0;
-
   color: white;
   text-align: center;
   font-weight: 600;
@@ -23,7 +17,7 @@
 }
 
 .vote-result-chart--slim .vote-result-chart__bars {
-  min-height: 6px;
+  min-height: 10px;
 }
 
 .vote-result-chart--slim .vote-result-chart__bar {
--- a/frontend/src/components/VoteResultChart.css
+++ b/frontend/src/components/VoteResultChart.css
@@ -24,9 +24,20 @@
   padding: 3px 0;
 }
 
-.vote-result-chart--slim .vote-result-chart__thumb,
-.vote-result-chart--slim .vote-result-chart__percentage {
+.vote-result-chart--slim .vote-result-chart__text {
   display: none;
+}
+
+.vote-result-chart__text {
+  padding: 2px;
+}
+
+.vote-result-chart__bar--for .vote-result-chart__text {
+  background: var(--green);
+}
+
+.vote-result-chart__bar--against .vote-result-chart__text {
+  background: var(--red);
 }
 
 .vote-result-chart__bar--for {
@@ -43,8 +54,7 @@
   background-color: var(--blue);
 }
 
-.vote-result-chart__bar--small .vote-result-chart__thumb,
-.vote-result-chart__bar--small .vote-result-chart__percentage {
+.vote-result-chart__bar--small .vote-result-chart__text {
   display: none;
 }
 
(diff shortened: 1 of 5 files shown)


frontend/src/components/VoteResultChart.css
```diff
@@ -1,5 +1,6 @@
 .vote-result-chart__bars {
   display: flex;
+  gap: 1px;
```
tillprochaska, line 3: The current 1px gap is achieved by setting `margin-left` a bit further down. I’d suggest using either `gap: 2px` or `margin-left: 2px` but not both at the same time!

tillprochaska:
> Also, I am unsure with the gaps - first question is if those even are the gaps that were meant?

Yes, I think so.

>  If those are the correct gaps - the result looks weird to me, but I think that will always be the case once we need to add space between them. Might just be a question of familiarization on my end.

It looks better if you remove the light gray background!

tillprochaska:
I’d probably try to add a solid green or red background behind the icon and percentage label -- otherwise the text can become a little difficult to read with the stronger background pattern.

frontend/src/components/VoteResultChart.css
linusha: Indeed. Do you remember why we did it with margin-left previously? I now opted for gap, as I find it easier to read?

linusha:
> It looks better if you remove the light gray background!

It does. I also removed the shadow, as that made weird lines through the empty space. However, I do not have a hard opinion on this. What do you think? I updated the picture above accordingly.

frontend/src/components/VoteResultChart.css
linusha: I also checked caniuse and support is good (no opera mini and old IE versions).

linusha:
> I’d probably try to add a solid green or red background behind the icon and percentage label -- otherwise the text can become a little difficult to read with the stronger background pattern.

Done.

linusha:
After looking at it for some time, I tend to think we should either make the small bars just a bit less tall or the large bar even taller? The difference in height is not as drastic now...

frontend/src/components/VoteResultChart.css
```diff
@@ -35,6 +29,18 @@
...
+  padding: 2px;
+}
+
+.vote-result-chart__text--for {
+  background: var(--green);
+}
+
+.vote-result-chart__text--against {
+  background: var(--red);
+}
```
tillprochaska, line 42: You can avoid adding extra classes to the markup if you use selectors like this:

```suggestion
.vote-result-chart__bar--for .vote-result-chart__text {
  background: var(--green);
}

.vote-result-chart__bar--against .vote-result-chart__text {
  background: var(--red);
}
```

frontend/src/components/VoteResultChart.css
```diff
@@ -35,6 +29,18 @@
   display: none;
```
tillprochaska, line 29: This is just details, but as the thumb and percentage are now wrapped in a single parent element, you can hide that as a whole. This fixes the case where both the thumb and the percentage are hidden, but you can still see the solid background:

Screen Shot 2024-08-30 at 16 23 01

There’s another instance of that at the end of the file for medium sized bars.

frontend/src/components/VoteResultChart.css
tillprochaska: Makes sense. `gap` support for flexbox elements has been worse a few years ago, so probably that was the reason.

tillprochaska:
>  I also removed the shadow, as that made weird lines through the empty space. However, I do not have a hard opinion on this. What do you think?

I think it looks good the way you implemented it!

tillprochaska:
> After looking at it for some time, I tend to think we should either make the small bars just a bit less tall or the large bar even taller? The difference in height is not as drastic now...

Agreed. Haven’t pulled the branch, but based on the screenshot I’m pretty sure the bar is larger than 10px. Maybe 

10-12px should look something like this:

Screen Shot 2024-08-30 at 16 31 51

Maybe this is due to the [empty text element](https://github.com/HowTheyVote/howtheyvote/pull/1024/files#r1738777271)?

frontend/src/components/VoteResultChart.css
linusha: Hah, good that you caught that. I actually did not see that :O :eyeglasses: 
Will fix and also see how to get the height to the intended value!

tillprochaska (approved):
This looks good to me now -- there’s just one small issue with the medium bars (see comment below), but feel free to merge after that’s resolved :)

PR #1025, opened 2024-08-27
Sort groups and countries alphabetically instead of implicitly by size

linusha:
When I filtered a bunch in order to prepare some slides for our talk, I found that not having the lists sorted by alphabet is a bit stressful. This makes it way easier in my opinion to find the country one is looking for. I have basically sorted the group list for the sake of consistency. The sorting in the position dropdown makes sense as is in my opinion, and also we know for sure that these options will never change, which is (more or less in theory) possible for the other values.

### DIFF
--- a/frontend/src/components/CountriesFilterSelect.tsx
+++ b/frontend/src/components/CountriesFilterSelect.tsx
@@ -14,7 +14,9 @@ export default function CountriesFilterSelect({
   onChange,
 }: CountriesFilterSelectProps) {
   const options = Object.fromEntries(
-    countries.map((country) => [country.code, country.label]),
+    countries
+      .map((country) => [country.code, country.label])
+      .sort((a, b) => a[1].localeCompare(b[1])),
   );
 
   return (
--- a/frontend/src/components/GroupsFilterSelect.tsx
+++ b/frontend/src/components/GroupsFilterSelect.tsx
@@ -14,7 +14,9 @@ export default function GroupsFilterSelect({
   onChange,
 }: GroupsFilterSelectProps) {
   const options = Object.fromEntries(
-    groups.map((group) => [group.code, group.short_label || group.label]),
+    groups
+      .map((group) => [group.code, group.short_label || group.label])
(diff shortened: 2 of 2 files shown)


frontend/src/components/CountriesFilterSelect.tsx
```diff
@@ -14,7 +14,7 @@ export default function CountriesFilterSelect({
   onChange,
 }: CountriesFilterSelectProps) {
   const options = Object.fromEntries(
-    countries.map((country) => [country.code, country.label]),
+    countries.map((country) => [country.code, country.label]).sort(),
```
tillprochaska, line 17:
```suggestion
    countries
      .map((country) => [country.code, country.label])
      .sort((a, b) => a[1].localeCompare(b[1])),
```

This doesn’t work reliably as it’s sorting by country code (which is different from the label that’s visible to users). For example, "Germany" would be listed before "Denmark". Instead, simply sort explicitly by label:

frontend/src/components/GroupsFilterSelect.tsx
```diff
@@ -14,7 +14,9 @@ export default function GroupsFilterSelect({
   onChange,
 }: GroupsFilterSelectProps) {
   const options = Object.fromEntries(
-    groups.map((group) => [group.code, group.short_label || group.label]),
+    groups
+      .map((group) => [group.code, group.short_label || group.label])
+      .sort(),
```
tillprochaska, line 19: Same as above, it’s better to sort by the visible label (although in case of the current groups this is the same as sorting by code):

```suggestion
    groups
      .map((group) => [group.code, group.short_label || group.label])
      .sort((a, b) => a[1].localeCompare(b[1])),
```

tillprochaska (approved):
I think changing the order is a good idea! Left comments below because the current implementation isn’t working reliably in all cases.

frontend/src/components/CountriesFilterSelect.tsx
linusha: Uff, yes. My bad. :disappointed:  Will fix and merge!

PR #1040, opened 2024-09-21
Remove superfluous white-space when filtering members

linusha:


frontend/src/components/MemberVotesList.test.tsx
```diff
@@ -90,6 +90,24 @@ describe("MemberVotesList", () => {
...
+      />,
+    );
+    const all = getAllByRole("listitem");
+    assert.strictEqual(all.length, 3);
+
+    await userEvent.type(getByRole("searchbox"), "max  muster");
+
+    const filtered = getAllByRole("listitem");
+    assert.strictEqual(filtered.length, 1);
+    assert.ok(within(filtered[0]).getByText("Max MUSTERMANN"));
```
tillprochaska, line 108: No need for a completely separate test case, but can you add an assertion below that tests that leading white space is ignored too?

linusha: Done :)

tillprochaska (approved):
Nice, thank you! :)

PR #1056, opened 2024-11-14
Update scraper after markup of source website changed

tillprochaska:
The markup of the OEIL has changed and the previous selector to extract the procedure title didn’t work any more. Also updated the tests fixture with the current version of the web page.


linusha (approved):
Have you checked whether this change is also retroactively applied? Otherwise, I still think we can merge this for now but we should keep note of that in order to not break the possibility of scraping older votes?

tillprochaska:
That’s a good point I hadn’t explicitly considered, but I did check a few older procedures (including the procedure used in tests) and they all use the new markup. As far as I know, OEIL pages are actually dynamically generated, unlike the Doceo stuff were changes seem to apply to newly published documents only.

PR #1058, opened 2024-11-16
Do not display votes with empty titles

tillprochaska:
In some cases, we cannot extract any title (vote title or procedure title) for a given vote. In that case, we cannot display anything about this vote in the frontend, so we should probably just hide it.

After deployment run the following commands:

```
# Delete search index to make sure all votes (incl. those without a title) are removed
htv system delete-indexes

# Recreate the index
htv system upgrade

# Reindex all votes
htv aggregate votes
```

### DIFF
--- a/backend/howtheyvote/api/votes_api.py
+++ b/backend/howtheyvote/api/votes_api.py
@@ -6,7 +6,7 @@ from io import StringIO
 from typing import TypeVar
 
 from flask import Blueprint, Response, abort, jsonify, request
-from sqlalchemy import select
+from sqlalchemy import or_, select
 from structlog import get_logger
 
 from ..db import Session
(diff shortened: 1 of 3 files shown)


linusha (approved):
I'd be fine merging this regardless, but as we have that option: Could we have a monitoring for votes with empty titles? Just so that we have the ability to see when we scrape those or at least keep track of how many there are?

tillprochaska:
We have a data quality check for empty titles and also expose this as a Prometheus metrics so it should be possible to create an alert based on it in Grafana: https://github.com/HowTheyVote/howtheyvote/blob/main/backend/howtheyvote/analysis/votes.py#L194

But we might want to reconsider how exactly we’re doing this at some point.

PR #1060, opened 2024-11-27
Update main vote analyzer to also handle special vote titles 

tillprochaska:
Until now, we’ve decided whether a vote is a main vote solely based on the description of a vote. However, in rare cases a main vote may not have description and only a title (e.g. in case of the election of the Commission). I’ve updated the analyzer to handle votes based on title in addition to the existing heuristic.


linusha:
Looks good! Danke!

PR #1080, opened 2025-01-13
Fix tests running against live data sources

linusha:
Our test-MEP excluded his birth date from his live profile page, resulting in the test against live data to fail. Hence, we have a new test-MEP.

As we provide a shortened version for the data-unchanged test-case in order to not create all necessary member entries,  we always need to mock this.

Some endpoints of the EP API (here the meetings endpoint) are at times quite slow. This makes it so that we  wait a bit longer for the poor hamster running their server.


tillprochaska (approved):
Nice, feel free to merge after fixing the linting errors :)

Issue #960, opened 2024-06-16
Required changes for new Parliamentary term


tillprochaska:
This is a WIP issue collecting required changes before the constituent sitting of the new Parliament in July:

- [x] Update `CURRENT_TERM` in [`config.py`](https://github.com/HowTheyVote/howtheyvote/blob/main/backend/howtheyvote/config.py#L28C1-L28C13)
- [x] Check that dates of upcoming plenaries have been scraped correctly.
- [x] Update list of political groups with `htv dev load-groups` some time before the constituent sitting. ("Political groups can be formed at any time during the legislative term, but they have an interest in being recognised before the constituent session (16 July)." – [source](https://epthinktank.eu/2024/04/29/timeline-to-new-eu-institutional-leadership/))
- [x] Update expected total number of MEPs in [`analysis.votes`](https://github.com/HowTheyVote/howtheyvote/blob/main/backend/howtheyvote/analysis/votes.py#L202).

linusha:
# To Do

- [ ] check if newly formed groups are incorporated into official data source
- [ ] #983

tillprochaska:
A new version of the "Corporate bodies" dataset has been published which includes ESN and PfE: https://op.europa.eu/en/web/eu-vocabularies/concept-scheme/-/resource?uri=http://publications.europa.eu/resource/authority/corporate-body

PR #1082, opened 2025-01-18
Update groups, countries, Eurovoc terms

tillprochaska:
This fixes the last remaining task in #960.

I’ve also implemented a few small changes to sort records and list values. This should make it much easier to diff the resulting JSON files. (Right now, rerunning some of the `htv load-*` commands can lead to large diffs, even though there have been only small or no changes.)

Also had to fix how we send SPARQL requests because some SPARQL endpoints now only accept POST requests. (We were using GET requests before.)

### DIFF
--- a/backend/howtheyvote/data/eurovoc.json
+++ b/backend/howtheyvote/data/eurovoc.json
@@ -3029,21 +3029,10 @@
   },
   {
     "id": "12",
-    "label": "accession to the European Union",
-    "alt_labels": [
-      "EU accession",
-      "accession to the Community",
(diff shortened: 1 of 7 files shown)


backend/howtheyvote/data/eurovoc.json
tillprochaska, line 1: The issue you were seeing should now be fixed. We should probably reaggregate votes after deployment using `htv aggregate votes` to make sure these changes are also reflected in the search index.

linusha (approved):

PR #1095, opened 2025-02-03
Remove "Featured" indicator in frontend

linusha:
We still use this under the hood, especially to inform the ranking. However, as we got quite a bit of feedback that it is not clear what this actually means, we remove it from the frontend for now.
We briefly discussed whether to rename it to something more telling, but decided against it, as our goal should be to get press releases/additional information for as many votes as possible anyways.

### DIFF
--- a/frontend/src/components/Tag.tsx
+++ b/frontend/src/components/Tag.tsx
@@ -1,15 +1,13 @@
 import type { ComponentChildren } from "preact";
-import { bem } from "../lib/bem";
 
 import "./Tag.css";
 
 type TagProps = {
-  style?: "featured";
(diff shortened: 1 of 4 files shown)


frontend/src/components/Tag.tsx
```diff
@@ -4,7 +4,7 @@ import { bem } from "../lib/bem";
 import "./Tag.css";
 
 type TagProps = {
-  style?: "featured";
+  style?: "";
```
tillprochaska, line 7: Simply remove the property entirely

tillprochaska (approved):

PR #1100, opened 2025-02-04
Add linux permissions fix for docker to README

### LINKED ISSUE #988: Fix permissions problem with files created inside of container on Linux
"My" problem is persisting even with a new and shiny installation of docker. The problem is described [here](https://stackoverflow.com/questions/70512898/docker-volume-mariadb-has-root-permission) as well and a solution is provided. Will fix this soon.

Fixing this will also resolve my problem of the read only db, which was not caused by another process locking a handle but was just a permissions problem due to the user inside of the container being root.

Fun fact: The reason why you never run into this @tillprochaska is that docker for mac resolves the underlying cause automagically.

linusha:
Resolves #988.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,11 @@ You can now access the application at `https://localhost`. Note that we use self
 
 At this point, there isn’t yet any data in your local database. You can use the [CLI](#cli) to run data pipelines.
 
+> [!TIP]
+> On Linux systems, due to the way Docker works by default, files created within the containers (such as the database file), will be owned by `root`.
+> In order to easily work with them via your ordinary user account, you can use the solution described [here](https://stackoverflow.com/a/70613576/4418325).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,6 +35,11 @@ You can now access the application at `https://localhost`. Note that we use self
 
 At this point, there isn’t yet any data in your local database. You can use the [CLI](#cli) to run data pipelines.
 
+> [!TIP]
+> On Linux systems, due to the way docker works by default, files created within the containers (such as the database file), will be owned by `root`.
```
tillprochaska, line 39:
```suggestion
> On Linux systems, due to the way Docker works by default, files created within the containers (such as the database file), will be owned by `root`.
```

tillprochaska (approved):

PR #1106, opened 2025-02-16
Scrape procedure committee

tillprochaska:
This PR includes changes to scrape the committee responsible for the legislative procedure a vote belongs to. We already scrape information about the procedure from the OEIL, so I simply extended what information the existing scraper extracts.

Similar to countries and political groups, I’ve loaded a list of all committees using the SPARQL endpoint of the Publications Office.

I’ve added the responsible committee to API responses and the CSV export, but for now it’s not displayed in the UI or available as a filter etc.

### DIFF
--- a/backend/howtheyvote/alembic/versions/3895af031aea_add_responsible_committee_column_to_.py
+++ b/backend/howtheyvote/alembic/versions/3895af031aea_add_responsible_committee_column_to_.py
@@ -1,0 +1,24 @@
+"""Add responsible_committee column to votes table
+
+Revision ID: 3895af031aea
+Revises: 2f958a6f147d
+Create Date: 2025-02-21 11:31:59.044124
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "3895af031aea"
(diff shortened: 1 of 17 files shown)


linusha:
Without the disabled package mode, I am not able to rebuild the `backend` container. However, I am also not able reproduce the issue you described in your commit message - I locally dropped the commit that removed that line from the `.toml` file, rebuild the container and afterwards could perfectly execute all `htv` commands inside of the resulting container?! :O

tillprochaska:
> Without the disabled package mode, I am not able to rebuild the backend container. 

Thanks, I forgot about that!

> However, I am also not able reproduce the issue you described in your commit message - I locally dropped the commit that removed that line from the .toml file, rebuild the container and afterwards could perfectly execute all htv commands inside of the resulting container?!

You need to do all of the following, otherwise the "old" `htv` CLI will still be there, even after you’ve rebuilt the image.

* Delete `.venv`
* `docker compose build backend`
* `docker compose run --rm backend poetry install`
* `docker compose up -d backend`

I have now fixed this so that building the Docker image succeeds, even when package mode is enabled (based on https://github.com/python-poetry/poetry/issues/8662#issuecomment-1905050446).

PR #1108, opened 2025-02-25
Match press releases and votes based on vote result

tillprochaska:
We can currently only match press releases with votes if the press release references a report or procedure. For non-legislative procedures, this is usually not the case. However, even press releases about non-legislative procedures usually mention the vote results (how many MEPs voted for/against/abstained). If there is exactly one vote with the same result on the day of the press release, we can use that info to match the press release to the vote.

I’ve implemented this as follows:

* As a first step, we now scrape the full text of a press release and try to extract vote results ("the text was adopted with x votes in favour, y against, and z abstentions" and similar wordings).

* Previously, we matched press releases to votes "on the fly" based on document and procedure references (which is fairly simple). Now, as the matching is getting a bit more complex, we do the matching as part of the `PressReleasesPipeline` and store the result in the database. Every vote now has a reference to the corresponding press release (if there is one).

* In order to match press releases to votes, we take document and procedure references (as before) as well as the vote results into account (new).

**Post deployment:**
* Run `htv press-releases` to re-scrape and match all press releases.
* Run `htv aggregate votes`

**To do:**

- [x] Scrape full text of press releases
- [x] Extract mentioned vote results from press release text
- [x] Persist matching of press releases and votes to database
- [x] Match press releases and votes based on vote result
- [x] Add CLI command to rerun press release scraper and analyzer for all existing press releases and votes
- [ ] ~Remove `is_featured` from database?~

### DIFF
--- a/backend/howtheyvote/pipelines/press.py
+++ b/backend/howtheyvote/pipelines/press.py
@@ -4,7 +4,7 @@ from collections.abc import Iterator
 from sqlalchemy import func, select
 from structlog import get_logger
 
-from ..analysis import FeaturedVotesAnalyzer
+from ..analysis import FeaturedVotesAnalyzer, VotePositionCountsAnalyzer
 from ..db import Session
 from ..models import PlenarySession, PressRelease, Vote
 from ..query import session_is_current_at
@@ -43,6 +43,7 @@ class PressPipeline(BasePipeline):
 
         self._scrape_press_releases_index()
         self._scrape_press_releases()
+        self._analyze_vote_position_counts()
         self._analyze_featured_votes()
         self._index_press_releases()
         self._index_votes()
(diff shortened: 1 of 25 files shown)


backend/howtheyvote/pipelines/press.py
```diff
@@ -138,9 +156,9 @@ def _analyze_featured_votes_by_date(self, date: datetime.date) -> None:
             )
             return
 
-        log.info("Running featured votes analysis", date=date)
+        log.info("Matching press releases and votes", date=date)
         writer = BulkWriter()
-        analyzer = FeaturedVotesAnalyzer(votes=votes, press_releases=self._press_releases())
+        analyzer = PressReleaseAnalyzer(votes=votes, press_releases=self._press_releases())
```
tillprochaska, line 161: Filter press releases by date

backend/howtheyvote/store/mappings.py
```diff
@@ -62,6 +62,9 @@ def map_vote(record: CompositeRecord) -> Vote:
     eurovoc_concepts = {EurovocConcept[id_] for id_ in record.chain("eurovoc_concepts")}
     responsible_committee = Committee.get(record.first("responsible_committee"))
 
+    press_release = record.first("press_release")
+    is_featured = record.first("is_featured") or press_release is not None
```
tillprochaska, line 66: Not sure we still need the `is_featured` column in the future.

tillprochaska: Decided to do this in a separate PR so not to make the diff even larger than it already is

linusha (approved):

PR #1109, opened 2025-02-28
Fix installing Python dependencies in Dockerfile

tillprochaska:
Follow up to 9f243166c95f80c3252b231cd9e0a37baefe968a. I’ve also sneaked in a small change to reduce the number of Docker layers for the backend and frontend images.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -21,10 +21,9 @@ RUN pip install poetry
 
 WORKDIR /howtheyvote/backend
 
-# Copy dependencies required to install dependencies
-COPY pyproject.toml pyproject.toml
-COPY poetry.toml poetry.toml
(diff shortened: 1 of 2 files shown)


linusha:
I cannot look into this right now (and you do not need to feel obliged to do it either), but this seems to make it worse?
My backend container does not start anymore, fails with:
```
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "htv": executable file not found in $PATH: unknown
```

tillprochaska:
Hm, can’t reproduce that. Let me know what the error message is if the problem persists and you can’t figure it out.

tillprochaska:
Merging this as I’d like to build a release, but we can of course change it later if it continues to be an issue.

PR #1116, opened 2025-03-05
Respect specified sort option when paginating through results

linusha:
Currently, when going on the search page, specifying a non-default sorting order and afterwards clicking the arrow to navigate to the second page, the sorting option will not be respected, effectively making it impossible to paginate through results not ordered by relevance.

This PR fixes that.

### DIFF
--- a/frontend/src/pages/SearchPage.tsx
+++ b/frontend/src/pages/SearchPage.tsx
@@ -24,8 +24,10 @@ const SORT_PARAMS = {
   oldest: { sort_by: "timestamp", sort_order: "asc" },
 } as const;
 
+type VotesSearchSortOptions = "relevance" | "newest" | "oldest";
+
 type SearchPageData = VotesQueryResponse & {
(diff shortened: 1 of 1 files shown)


frontend/src/pages/SearchPage.tsx
```diff
@@ -68,13 +70,19 @@ export const loader: Loader<SearchPageData> = async (request: Request) => {
...
+  sort: VotesSearchSortOptions,
+  page: number,
+): string {
   const params = new URLSearchParams();
 
   if (query !== "") {
     params.set("q", query);
   }
 
+  params.set("sort", sort);
```
tillprochaska, line 84: Maybe only add the parameter when it’s not the default?

```suggestion
  if (sort !== DEFAULT_SORT) {
    params.set("sort", sort);
  }
```

tillprochaska (approved):
Ah, that’s a very good point!

PR #1120, opened 2025-03-14
Add EuroVoc concepts, geographic areas to CSV export

### LINKED ISSUE #1117: Remove `isFeatured` column from database
See #1108.

tillprochaska:
This PR includes three changes to the CSV export:

**Remove `is_featured` column**

Discussed before, closes #1117. This removes the column everywhere (including from the API and exports). Whether a vote has a corresponding press release is still a ranking factor, but it’s now named `has_press_release` in the search index which should be more obvious (but isn’t exposed publicly).

**Include EuroVoc concepts**
While exposed via the API, we haven’t included the EuroVoc concepts in the export so far. This adds `eurovoc_concepts.csv` and `eurovoc_concept_votes.csv` to the export.

**Include geographic areas**
We already have `countries.csv` which includes data about the EU member states and is referenced by the MEPs export. This adds a separate table `geo_areas.csv` for the geographic areas referenced by votes. I’m not really sure about this being a separate table though. Maybe we should just have a single reference table `geo_areas.csv` that is referenced by the MEPs export and the votes export?

### DIFF
--- a/backend/howtheyvote/store/index.py
+++ b/backend/howtheyvote/store/index.py
@@ -134,16 +134,12 @@ def _serialize_vote(vote: Vote, generator: TermGenerator) -> Document:
     if vote.rapporteur:
         generator.index_text(vote.rapporteur, 1, field_to_prefix("rapporteur"))
 
-    # Store timestamp and is_featured as sortable values for ranking
+    # Store timestamp and press release as sortable values for ranking
     timestamp = sortable_serialise(vote.timestamp.timestamp())
     doc.add_value(field_to_slot("timestamp"), timestamp)
 
-    is_featured = sortable_serialise(int(vote.is_featured))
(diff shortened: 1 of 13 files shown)


backend/howtheyvote/store/index.py
```diff
@@ -127,23 +127,19 @@ def _serialize_vote(vote: Vote, generator: TermGenerator) -> Document:
 
     # Index geographic areas for full-text search
     for geo_area in vote.geo_areas:
-        generator.index_text(geo_area.label, 1, field_to_prefix("geo_areas"))
+        generator.index_text(geo_area.code, 1, field_to_prefix("geo_areas"))
```
tillprochaska, line 130: Note to self: I don’t think this is what we want

PR #1122, opened 2025-03-16
Handle votes with joint responsible committees

### LINKED ISSUE #1121: There can be more than one responsible committee
For example https://oeil.secure.europarl.europa.eu/oeil/en/procedure-file?reference=2024/0258(COD)

tillprochaska:
A procedure can have multiple responsible committees (see [Rules of Procedure](https://www.europarl.europa.eu/doceo/document/RULES-9-2022-07-11-RULE-058_EN.html) and an [example procedure](https://oeil.secure.europarl.europa.eu/oeil/en/procedure-file?reference=2024/0258(COD))).

* I’ve updated the procedure scraper to properly handle procedures with a joint committee responsible and added tests.
* I’ve updated the CSV export and removed the `responsible_committee_code` in the `votes.csv` table. Instead, there are now separate tables `committees.csv` and `responsible_committee_votes.csv`.

After deployment, we need to re-scrape the procedure files and then re-aggregate the votes.

Fixes #1121

### DIFF
--- a/backend/howtheyvote/export/__init__.py
+++ b/backend/howtheyvote/export/__init__.py
@@ -9,7 +9,7 @@ from structlog import get_logger
 
 from ..db import Session
 from ..helpers import parse_procedure_reference
-from ..models import Country, EurovocConcept, Member, Vote
+from ..models import Committee, Country, EurovocConcept, Member, Vote
 from ..vote_stats import count_vote_positions
 from .csvw_helpers import Table
 
@@ -149,9 +149,6 @@ class VoteRow(TypedDict):
     `OLP_SECOND_READING`, `OLP_THIRD_READING`.This field is only available for votes starting
     in 2024 and if the vote is part of an Ordinary Legislative Procedure."""
 
-    responsible_committee_code: str | None
-    """Committee responsible for the legislative procedure"""
-
     count_for: int
     """Number of MEPs who voted in favor"""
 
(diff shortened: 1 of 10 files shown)


linusha (commented):
More of a conceptual objection: I think extracting the responsible committees from the votes.csv severely limits the usefulness of that file and makes finding stuff on specific topics (or maybe better: limiting the list of votes to sensible candidates) much harder.
I know that one can *just join* the two tables, but I am thinking about a quick Excel-thing here. 

There are multiple ways to handle this imo:
1. One can say does not matter and just do it anyways, and rethink this once negative feedback comes in. I think relying on this is not a good solution, in this case.
2. I would need to spend some more time on reading the procedural rules, but as long as there is a hard cap on the number of responsible committees, we could just include several columns in the votes.csv, most of them empty.
3. I even would be okay with a solution that just has one columns and string-concatenates them ("LIBE, IMCO", as an example).

Side-note: As its cheap, should we also collect the Committees for opinion? Probably should not even be in this PR and here I would also be totally fine with extracting that in a separate table in any case. Just wanted to raise the matter here as it seemed fitting.

backend/howtheyvote/export/__init__.py
```diff
@@ -182,6 +179,29 @@ class MemberVoteRow(TypedDict):
...
+    """Unique identifier of the committee"""
+
+    label: str
+    """Label"""
+
+    abbreviation: str
+    """Abbreviation"""
+
+
+class ResponsibleCommitteeVotes(TypedDict):
```
tillprochaska, line 195: Should have `Row` suffix

backend/howtheyvote/export/__init__.py
```diff
@@ -554,6 +595,26 @@ def export_geo_areas(self) -> None:
...
+
+            for committee_code in committee_codes:
+                committee = Committee[committee_code] if True else None
+                committees.write_row(
+                    {
+                        "code": committee.code,
+                        "label": committee.label,
+                        "abbreviation": committee.abbreviation,
+                    }
+                )
```
linusha, line 616: Just for my understanding - this means that the committee table will not be guaranteed to contain all committes, but all committees that were responsible for a vote that we collected, correct?

linusha (approved):
I still think the way we separate the tables here now at least point towards an inconsistency in our mental-model of possible target audiences for the export.

I don't think spending time or energy on resolving this makes sense. Can be merged.

backend/howtheyvote/export/__init__.py
tillprochaska: Yes, exactly. The list of committees we obtain from the Publications Office contains every committee that has ever exists (some of them which have existed for only a very short amount of time decades ago). This is similar to what we do for e.g. EuroVoc concepts as well, we only export what’s actually relevant to the dataset.

PR #1125, opened 2025-03-16
Scrape English RCV lists

tillprochaska:
- [ ] Set up worker
- [x] Remove suffix for procedure stage (e.g. "***I)
- [ ] Ensure that the language of the returned document is English, raise DataNotAvailableException otherwise


tillprochaska:
Closing this for now. Even when requesting the English version of the RCV lists, the French version is returned if the English translations aren’t yet available. While the root XML tag has a `lang` attribute, it doesn’t relate to the actual language of the document contents, and there doesn’t seem to be another way to check whether the contents are in English or French (besides running language detection).

We’ll soon start scraping the VOT table, which also include the English titles (besides other data). The VOT tables seem to be published with a bit more delay than the English RCV lists, but that seems like a fair trade-off.

PR #1131, opened 2025-04-11
Add VOT documents as new data source

tillprochaska:
- [x] Add new pipeline to scrape VOT lists
- [ ] Set up scheduled worker task
- [ ] Add amendment info (subject, number, author)
- [x] Derive procedure stage from title
- [x] Add new data to API
- [x] Add new data to export


backend/howtheyvote/models/vote.py
```diff
@@ -84,13 +90,15 @@ class Vote(BaseWithId):
     term: Mapped[int] = mapped_column(sa.Integer)
     order: Mapped[int] = mapped_column(sa.Integer)
     title: Mapped[str | None] = mapped_column(sa.Unicode)
+    dlv_title: Mapped[str | None] = mapped_column(sa.Unicode)
```
tillprochaska, line 99: DLV seems to be the abbreviation for a system the EP uses for this data sources. The name is not great, but I couldn’t come up with something else, and it isn’t exposed in the export and API.

backend/howtheyvote/api/serializers.py
```diff
@@ -24,6 +26,10 @@ class ProcedureDict(TypedDict):
     reference: Annotated[str, "2022/0195(COD)"]
     """Procedure reference as listed in the Legislative Observatory"""
 
+    stage: ProcedureStage | None
+    """Stage of the procedure at the time of the vote. This field is only available for votes
```
linusha, line 30:
```suggestion
    """Stage of the procedure in which the vote took place. This field is only available for votes
```

backend/howtheyvote/export/__init__.py
```diff
@@ -139,6 +139,11 @@ class VoteRow(TypedDict):
     procedure_title: str | None
     """Title of the legislative procedure as listed in the Legislative Observatory"""
 
+    procedure_stage: str | None
+    """Stage of the procedure at the time of the vote. One of `OLP_FIRST_READING`,
```
linusha, line 143: Keep this consistent with the one above

backend/howtheyvote/models/vote.py
tillprochaska: For future reference: Each block/group of votes has a DLV ID, and this is the title of the block. It might also be an abbreviation for something else, not the IT system.

linusha (approved):

PR #1134, opened 2025-04-22
Add procedure type to API and export

tillprochaska:
This adds the procedure type as a separate field to the API and CSV export. As the procedure type is a part of the procedure reference (e.g. "COD" in "1234/2025(COD)"), technically, this is already part of API responses and CSV exports. However, in practice, it’s quite convenient to have this exposed as a separate field for filtering etc.


tillprochaska:
Merging this now as it’s a rather small change and is otherwise causing conflicts with the work I’m doing right now to fetch amendmend info (authors, …).

PR #1160, opened 2025-06-08
Send Pushover notifications for important events

tillprochaska:
This uses Pushover to send notifications for important events:

* New vote results are available.
* A pipeline fails. (If data isn't available yet, that isn't considered a pipeline failure and no notification is sent.)
* Generating sharepics fails. (This sends only one notification per pipeline run, even if multiple sharepics couldn’t be generated.)

In the future, I’d like to also implement notifications when we attempt to scrape a pipeline data source for the last time and data still isn’t available.

To test this:

1. Sign up for a [Pushover](https://pushover.net/) account
2. Create a new [API application](https://support.pushover.net/i175-how-to-get-a-pushover-api-or-pushover-application-token)
3. Add the API token and your user key (from the Pushover dashboard) to your `.env` file
4. Recreate the backend container: `docker compose up -d backend`
5. Run the RCV list pipeline from within the `backend` container: `htv pipeline rcv-list ...`

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -1,3 +1,7 @@
 HTV_BACKEND_PUBLIC_URL=https://localhost/api
 HTV_FRONTEND_PUBLIC_URL=https://localhost
 CADDY_SITE_ADDRESS=localhost
+
+# Optional
+PUSHOVER_API_TOKEN=
+PUSHOVER_USER_KEY=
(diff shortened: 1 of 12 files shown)


linusha (approved):
TODO: Add tokens to ansible vault in /infra

PR #1168, opened 2025-06-09
Check language of VOT document

### LINKED ISSUE #1167: Check whether VOT lists are guaranteed to be English
I vaguely remember that requesting a VOT list like [this](https://www.europarl.europa.eu/doceo/document/PV-10-2025-05-22-VOT_EN.xml) may occasionally return the French version if the English translation isn’t ready yet (instead of returning 404 🥺). Usually, the English translation is published with a few days after the French original has been published.

While the XML version has a `language` attribute on the root element, if I remember correctly, this is set to `EN` if the requested language is English, even if the actual contents are French.

What we need to do:

* Check whether the language attribute is really returning `EN` even if the contents are French. This has to be done during a session. Next opportunity is June 16th.
  * If yes, add a check to the `VOTListScraper` and raise an exception if the French version is returned
  * If no, in the `VOTListScraper`, run some sort of language detection on the titles from the VOT list, and raise an exception if it is French.

tillprochaska:
If an English translation isn’t yet available, requesting the English translation will return the French original. This checks whether the returned document actually is the English version. If not, it raises a `NoWorkingUrlError` exception, which will be treated in the same way as if the EP website had returned a 404 (i.e. the scraper will be retried later).

Fixes #1167

This is a draft PR for now, because I do need to verify whether the returned document actually includes the correct language code. See issue linked above for details.


tillprochaska:
Ok, checked this today. When the English translations aren’t available, the French original is returned. In that case, the `language` attribute is accurate, i.e. it is set to `FR`. So this solution should be enough, no need for additional language detection :)

* https://web.archive.org/web/20250617170722/https://www.europarl.europa.eu/doceo/document/PV-10-2025-06-17-VOT_FR.xml
* https://web.archive.org/web/20250617170635/https://www.europarl.europa.eu/doceo/document/PV-10-2025-06-17-VOT_EN.xml

linusha (approved):

PR #1185, opened 2025-07-11
Add scheduling for VOT pipelines

tillprochaska:
VOT lists currently have to be run manually. In contrast to RCV lists, VOT lists are usually published with at least one day of delay. That means we have to retry fetching VOT lists for multiple days until the data becomes available.

To solve this, I’ve added "idempotency keys" to scheduled pipelines. If such a key is given, a pipeline will only be executed until it has been executed successfully for the first time.

As the existing scheduling logic for RCV pipelines was mostly untested, I’ve added test coverage and then refactored it to also make use of idempotency keys.

Finally, I’ve added the actual scheduling of VOT lists.

Most of the diff is due to new tests, the number of lines added for the idempotency keys is fairly small. It works quite well for the RCV use case, but I think the scheduling of the VOT lists is a bit unintuitive. Not sure what a better/simpler solution would be though.

I’d recommend going through the changes commit by commit.

### DIFF
--- a/backend/howtheyvote/worker/__init__.py
+++ b/backend/howtheyvote/worker/__init__.py
@@ -29,7 +29,7 @@ from .worker import (
 log = get_logger(__name__)
 
 
-def op_rcv() -> PipelineResult:
+def rcv_list_handler() -> PipelineResult:
     """Checks if there is a current plenary session and, if yes, fetches the latest roll-call
     vote results."""
     today = datetime.date.today()
@@ -51,43 +51,9 @@ def op_rcv() -> PipelineResult:
     return pipeline.run()
 
 
-def op_press() -> PipelineResult:
-    """Checks if there is a current plenary session and, if yes, fetches the latest press
-    releases from the Parliament’s news hub."""
-    today = datetime.date.today()
-
-    if not _is_session_day(today):
-        raise SkipPipeline()
-
-    pipeline = PressPipeline(date=today, with_rss=True)
-    return pipeline.run()
-
-
-def op_sessions() -> PipelineResult:
-    """Fetches plenary session dates."""
-    pipeline = SessionsPipeline(term=config.CURRENT_TERM)
-    return pipeline.run()
-
-
-def op_members() -> PipelineResult:
-    """Fetches information about all members of the current term."""
-    pipeline = MembersPipeline(term=config.CURRENT_TERM)
-    return pipeline.run()
-
-
-EXPORT_LAST_RUN = Gauge(
-    "htv_export_last_run_timestamp_seconds",
-    "Timestamp when the CSV export was generated the last time",
-)
-
-
-def op_generate_export() -> None:
-    archive_path = file_path("export/export")
-    generate_export(archive_path)
-    EXPORT_LAST_RUN.set(time.time())
-
-
-def op_notify_last_run_unsuccessful() -> None:
+def rcv_list_notification_handler() -> None:
+    """Checks whether the RCV list pipeline has been executed successfuly, and sends a
+    Pushover notification if not."""
     today = datetime.date.today()
 
     # Do not check for last run on days without Plenary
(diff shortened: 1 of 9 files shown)


backend/howtheyvote/worker/worker.py
```diff
@@ -175,13 +175,31 @@ def schedule_pipeline(
         hours: Iterable[int] = {0},
         minutes: Iterable[int] = {0},
         tz: str | None = None,
+        idempotency_key: Callable[..., str] | None = None,
```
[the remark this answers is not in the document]
tillprochaska: Although maybe being explicit here is better…

backend/tests/worker/test_init.py
tillprochaska, line 1: The tests in this file aren’t great to read/follow due to the amount of mocking involved, not sure if there’s a better solution…

backend/howtheyvote/worker/__init__.py
```diff
@@ -101,98 +131,106 @@ def op_members() -> PipelineResult:
...
+    for i in range(1, 8):
+        worker.schedule_pipeline(
+            make_vot_list_handler(i),
+            name=VOTListPipeline.__name__,
+            hours={21},
+            tz=config.TIMEZONE,
+            idempotency_key=(
+                lambda i=i: (datetime.date.today() - datetime.timedelta(days=i)).isoformat()
+            ),
+        )
```
tillprochaska, line 210: This works and it’s just a few LOC, but it’s still quite a bit of complexity…

backend/howtheyvote/worker/worker.py
```diff
@@ -175,13 +175,31 @@ def schedule_pipeline(
...
+            idempotency_key_value = None
+
+            # Check if an idempotency key function is given, and skip this run if there was a
+            # previous successful run of the pipeline with the same key.
+            if idempotency_key:
+                idempotency_key_value = idempotency_key()
+                count = (
+                    select(func.count())
+                    .select_from(PipelineRun)
+                    .where(PipelineRun.idempotency_key == idempotency_key_value)
```
linusha, line 181: As you said in the comment above yourself, I think it would be nicer to make the naming more explicit. Especially for the `Callable`, something like `idempotency_key_function` would already help a lot I think, does not need to be sophisticated. Otherwise, it becomes really confusing, especially with something like line 195. Here, we could also think about renaming the property on the `PipelineRun`, but that's of lower importance IMO.

backend/tests/worker/test_init.py
```diff
@@ -0,0 +1,202 @@
...
+
+        worker.run_pending()
+
+        # The pipeline has been executed again, as it’s the next day
+        runs = list(db_session.execute(query).scalars())
+        assert len(runs) == 5
+        assert runs[4].status == PipelineStatus.SUCCESS
+
+
+def test_rcv_list_notification(db_session, mocker):
```
linusha, line 137: Can we rename this test? In the case we successfully scrape new votes, we also send a notification. This is untested, as the actual behavior of the `RCVListPipeline` is mocked, if I am not mistaken.
Hence, this tests whether the mechanism to notify us when there were no votes, but we think there might should be votes. Something like `test_rcv_list_notification_no_votes` would suffice, mainly to make it more explicit what is (and is not) tested.

linusha (commented):
Thank you!
Overall, I found this okay to follow! Not sure how good of a reference I am, but even the tests are not too bad I think :slightly_smiling_face: 
I would like to address the minor naming things I raised. Otherwise :+1:

backend/howtheyvote/worker/__init__.py
```diff
@@ -24,54 +26,82 @@
...
 
 
-def op_rcv_midday() -> PipelineResult:
+def rcv_list_handler() -> PipelineResult:
     """Checks if there is a current plenary session and, if yes, fetches the latest roll-call
     vote results."""
     today = datetime.date.today()
 
     if not _is_session_day(today):
         raise SkipPipeline()
```
linusha, line 40: I just though that maybe we want to extract this check into the `schedule_pipeline` function? We have that in multiple places, basically everything that depends on plenary information. Would reduce redundancy, but then we would need a flag that enables/disables the check e.g., for the `MemberPipeline` - so not too sure how much we win there. 
No hard opinion here, just a thought.

backend/tests/worker/test_init.py
tillprochaska: @linusha I’ve added a separate test for the entire `RCVListPipeline` which also includes the success notification.

backend/howtheyvote/worker/__init__.py
tillprochaska: On one hand, we do indeed have quite a few pipelines with that check. On the other hand, the worker implementation is currently not concerned with details of our specific use case, and I still have hope that one day we’ll find something off-the-shelf that can replace our semi-custom worker… :)

PR #1186, opened 2025-07-20
Add section with external links about vote

tillprochaska:


backend/howtheyvote/api/votes_api.py
```diff
@@ -436,6 +441,39 @@ def _format_sources(fragments: Iterable[Fragment]) -> list[SourceDict]:
...
+                "description": "Original text of the report or resolution as tabled. The text may be different from the adopted text if MEPs have adopted amendments.",  #  noqa: E501
+                "url": doceo_document_url(vote.reference),
+            }
+        )
+
+    if vote.procedure_reference:
+        links.append(
+            {
+                "title": "Legislative Observatory",
+                "description": "Find out more about the procedure this vote is part of. This includes information about the current state of the procedure, past and upcoming steps, as well as key players.",  #  noqa: E501
```
linusha, line 469:
```suggestion
                "description": "Find out more about the legislative procedure this vote is part of. This includes information about the current state of the procedure, past and upcoming steps, as well as key players.",  #  noqa: E501
```

backend/howtheyvote/api/votes_api.py
```diff
@@ -436,6 +441,39 @@ def _format_sources(fragments: Iterable[Fragment]) -> list[SourceDict]:
...
+        )
+
+    if vote.reference:
+        links.append(
+            {
+                "title": "Report or resolution",
+                "description": "Original text of the report or resolution as tabled. The text may be different from the adopted text if MEPs have adopted amendments.",  #  noqa: E501
+                "url": doceo_document_url(vote.reference),
+            }
+        )
```
linusha, line 463: We do not need to do this now, but I think we should consider the following:
In the cases where amendments have been adopted, the resolution page will often (I would hope always, but who knows) display a link to the text as it was adopted in the "flow-chart" under "Texts adopted". Scraping that as well would of course increase complexity, but make contents of the link we provide far more intuitive.
The link is also retrievable from OEIL.

linusha (commented):
I am strongly in favor of moving the "More information" section further up the page.
The press-release extracts already provide some information about the contents of the vote, and I think it makes sense to group everything there.

In order for this to not take up too much space (especially as we integrate further additional resources), we can of course make these new contents collapsible.

Additionally, I think it would make sense to integrate the extracts from the press release and the new link. Especially, as we currently provide no information where the bullet points come from on the page. I'm absolutely open to other options, the first that came to mind would be something like this?

```
# Press Release
- Extracted
- Extracted
- Extracted

Press releases, published by the European Parliament’s Press Service, often contain more information about the subject of the vote and next steps. Read the [full press release] here.
```
Although that's still quite implicit to be honest...

backend/howtheyvote/api/votes_api.py
tillprochaska: I think initially I had included "legislative" here as well, but eventually removed it because the OEIL also covers stuff that isn’t legislative. (Not sure anyone but us is thinking about details like this, I’m fine with either option.)

tillprochaska:
> I am strongly in favor of moving the "More information" section further up the page.
> The press-release extracts already provide some information about the contents of the vote, and I think it makes sense to group everything there.

I get the point, but that would effectively mean puhsing down the vote results quite a lot. While I think that the external links are helpful, I don’t think they are as important as the actual vote results.

I have two suggestions to solve that problem:

1) Adding a ToC/nav element with links to the different sections of the page

Screenshot 2025-07-24 at 8 08 02 PM

2) Adding a small link below the bullet points which targets the "More information" section:

Screenshot 2025-07-24 at 8 09 39 PM

(I’m not set on that label, "Learn more", "More information", "Learn more about this vote", etc. would all be fine imo.)

backend/howtheyvote/api/votes_api.py
linusha: Then let's just go with the more correct option - sorry, I got duped by their name. :see_no_evil:

linusha:
> I get the point, but that would effectively mean puhsing down the vote results quite a lot. While I think that the external links are helpful, I don’t think they are as important as the actual vote results.

Agreed! I just didn't think far enough the reach the options you proposed. 

> I think I have a slight preference for the first option.

Me too.

backend/howtheyvote/scrapers/votes.py
linusha, line 1: I can't (or don't know how to) comment on commit messages, so here we go:

The rules of procedure state:
> Rule 209 - Texts adopted
> 1. Texts adopted by Parliament shall be published immediately after the vote. They shall be
> placed before Parliament in conjunction with the minutes of the relevant sitting and preserved in
> the records of Parliament.

So I think you are absolutely right with your assumption - and how it goes in reality we just need to see.

tillprochaska:
>> I think I have a slight preference for the first option.
>
> Me too.

@linusha Would it be ok to merge this PR and then I can open a separate PR to add the ToC as discussed above? (Unless there’s something else in this PR you’d like to see changed of course!)

PR #1207, opened 2025-09-18
Remove data issues

tillprochaska:
This was originally intended to perform continuous data quality checks. However, in practice, we don't really use this and rely a lot on manual checks. At the same time, this adds quite a bit of complexity to the code base, and adding new checks is more work than it should be.

I think a better way of running these checks would be in a separate system, for example in a GitHub Action based on our data exports.


backend/howtheyvote/pipelines/rcv_list.py
```diff
@@ -84,7 +84,7 @@ def _run(self) -> None:
         # Send Pushover notification
         send_notification(
             title=f"Scraped RCV list for {self.date.strftime('%a, %b %-d')}",
-            message=f"{len(self._vote_ids)} votes, {len(self._vote_group_ids)} vote groups",
+            message=f"{len(self._vote_ids)} votes",
```
linusha, line 83: > a more useful alternative would be to include the number of main votes (which is what would be listed on our website)

Quite often the number in the notification matched that exactly. Not every time, but often enough that I found it to be quite useful. But I am also in favor of making this explicit (and correct in all cases) this is just for informational purposes.

linusha (approved):
I am in favor of merging this, but I would open two new tickets:

1. Adding the number of new votes visible on the website (this time for real) to the notification. Possibly also the number of new different vote-group hashes (which ideally should be similar to the other number). This one is somewhat important.
2. As you said, automating some quality checks via an action. This is less important. I will just link to this PR in the new ticket, its easy to spot what we once had and then to consider whether this is still interesting to us or not.

PR #1210, opened 2025-09-20
Add CLI command to delete fragments

### LINKED ISSUE #1171: Add CLI command to delete fragments for a specific record by source

tillprochaska:
Pipelines/scrapers only insert or update fragments, but they never delete them. That means if a fragment has been written erroneously (either because of an error in the source data or because of a bug in the software) it has to be deleted explicitly.

This adds a CLI command to manually delete fragments. Usage examples:

```
htv flush votes --id 12345
htv flush votes --id 12345 --id 67890
htv flush votes --date 2025-09-20
htv flush votes --source-name RCVListScraper
htv flush votes --source-name RCVListScraper --id 12345
```

Fixes #1171

### DIFF
--- a/backend/howtheyvote/cli/__init__.py
+++ b/backend/howtheyvote/cli/__init__.py
@@ -11,6 +11,7 @@ from ..sharepics import generate_vote_sharepic
 from ..worker import worker as worker_
 from .aggregate import aggregate
 from .dev import dev
+from .flush import flush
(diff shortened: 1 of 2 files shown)


tillprochaska:
I’m going to merge this without a review because it would be super helpful for other features I’m working on at the moment, but I’d still be open to changes to the implementation or the CLI interface.

PR #1224, opened 2025-10-06
Filters UI

tillprochaska:
Merge #1212 and #1225 first.

This adds a basic filter UI to the search results page:

https://github.com/user-attachments/assets/4686b00c-3169-425c-8897-e1495ccd784e

There are a number of potential enhancements. Not all of them need to be a top priority though, and I think it would make sense to implement them in separate PRs to keep the scope of this PR manageable.

* On desktop, display the filters in a sidebar (rather than hiding them in a dialog window).
* Allow users to search within filter options (the "Countries" filter can have a lot of options).
* Show the active filters above the search results (also allowing users to remove specific or all filters).
* For the dates filter, provide a few default ranges (e.g. "Current plenary term", "Current year", …)
* Debug the date pickers in different browsers and fix some inconsistencies.
* ~~Shorten committee labels to remove the "Committee on the" prefix as it’s redundant.~~ Potentially also include the committee abbreviation in the label.
* When a user changes a filter selection, automatically refresh the options of other filters (e.g. when selecting "Germany", update the options of the Committees filter to include only options for the subset of votes matching "Germany").


tillprochaska:
Same for me, it ran successfully locally

PR #1242, opened 2025-10-24
Add amendment information and adopted text reference

linusha:
This should make the export contain all information currently available in the DB which we can re-share.

### DIFF
--- a/backend/howtheyvote/export/__init__.py
+++ b/backend/howtheyvote/export/__init__.py
@@ -182,7 +182,7 @@ class VoteRow(TypedDict):
     """Vote result. One of `ADOPTED`, `REJECTED`, `LAPSED`. This field is only available for
     votes starting in 2024."""
 
-    adopted_text_reference: str | None
+    texts_adopted_reference: str | None
     """"EP reference number of the text as it was adopted.
     This includes all amendments that were adopted to an original text."""
 
@@ -309,7 +309,7 @@ class AmendmentAuthorVoteRow(TypedDict):
     """Vote ID"""
 
     author_type: str
-    """Either GROUP, COMMITTEE, MEMBERS, ORALLY, ORIGINAL_TEXT, or RAPPORTEUR"""
+    """Either `GROUP, `COMMITTEE`, `MEMBERS`, `ORALLY`, `ORIGINAL_TEXT`, or `RAPPORTEUR`"""
 
     group_code: str
     """Group code, if applicable."""
@@ -424,7 +424,7 @@ class Export:
             row_type=AmendmentAuthorVoteRow,
             outdir=self.outdir,
             name="amendment_authors_votes",
-            primary_key=["vote_id"],
+            primary_key=["vote_id", "author_type", "group_code", "committee_code"],
         )
 
     def run(self) -> None:
@@ -598,7 +598,7 @@ class Export:
                         "count_abstention": position_counts["ABSTENTION"],
                         "count_did_not_vote": position_counts["DID_NOT_VOTE"],
                         "result": vote.result.value if vote.result else None,
-                        "adopted_text_reference": vote.texts_adopted_reference,
+                        "texts_adopted_reference": vote.texts_adopted_reference,
                     }
                 )
 
--- a/backend/tests/export/test_init.py
+++ b/backend/tests/export/test_init.py
@@ -149,7 +149,7 @@ def test_export_votes(db_session, tmp_path):
         result=VoteResult.ADOPTED,
         amendment_subject=None,
         amendment_number=None,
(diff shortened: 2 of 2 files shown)


backend/howtheyvote/export/__init__.py
```diff
@@ -534,9 +543,14 @@ def export_votes(self) -> None:
                         "timestamp": vote.timestamp,
                         "display_title": vote.display_title,
                         "reference": vote.reference,
+                        "amendment_number": vote.amendment_number,
+                        "amendment_authors": "; ".join(
+                            aa.stringify() for aa in (vote.amendment_authors or [])
+                        ),
```
tillprochaska: This probably boils down to the same discussion we had with regard to the responsible committees: I’d prefer a normalized approach for the export, i.e. a separate `amendment_authors` table that looks something like this:

| vote_id | type | group_code | committee_code |
| --- | --- | --- | --- |
| 1 | GROUP | PPE | |
| 2 | COMMITTEE | LIBE | |
| 3 | MEMBERS | | |
| … | … | … | … |

In my opinion, this would make it much easier to reliably handle use cases such as "Identify all votes on amendments that were tabled by political groups/a specific group" or "Count number of votes on amendments by author".

In some cases, doing simple substring filters may be enough, for example to identify all votes on amendments tabled by a particular group:

```py
# `amendment_authors` column
sd_amendments = votes_df[votes_df["authors"].str.contains("SD group")]]

# `amendment_authors` table
sd_authors = authors_df[(authors_df["type"] == "GROUP") & (authors_df["group_code"] == "SD")]
sd_amendments = votes[votes["id"].isin(sd_authors["vote_id"])]
```

(I haven’t actually run this code, so there might be errors.)

The second option obviously is more complex (although I’d argue it’s still relatively easy to understand). However, the first option requires that you’re absolutely sure that there can be no overlaps in, e.g., group codes. (Which I think is true at the moment, but you never know given the pace at which new far-right groups are founded.)

What if we want to count the number of amendment votes by the political group that tabled the amendment? Something like this should work (again untested) if we have a separate `amendment_authors` table:

```py
counts = (
    authors_df[authors_df["type"] == "GROUP"]
    .groupby("group_code")["vote_id"]
    .nunique()
)
```

I’m not entirely sure what the best option would be in case of a stringified `amendment_authors` column, but I think the following could work:

```py
votes_df["amendment_authors_list"] = votes_df["amendment_authors"].str.split(",")
authors = votes_df.explode("amendment_authors_list")
counts = (
    authors[authors["amendment_authors_list"].str.endswith("group")]
    .groupby("group_code")["vote_id"]
    .nunique()
)
```

In this case, I think the first option is less complex, but, more importantly, also much easier to understand.

backend/howtheyvote/export/__init__.py
```diff
@@ -280,6 +301,23 @@ class ResponsibleCommitteeVoteRow(TypedDict):
...
 
+class AmendmentAuthorVoteRow(TypedDict):
+    """Authors of the vote if vote is amendment. Multiple authors result in multiple rows.
+    This information is only available for votes starting in 2024."""
+
+    vote_id: int
+    """Vote ID"""
+
+    author_type: str
+    """Either GROUP, COMMITTEE, MEMBERS, ORALLY, ORIGINAL_TEXT, or RAPPORTEUR"""
```
tillprochaska:
```suggestion
    """Either `GROUP`, `COMMITTEE`, `MEMBERS`, `ORALLY`, `ORIGINAL_TEXT`, or `RAPPORTEUR`"""
```

backend/tests/export/test_init.py
```diff
@@ -147,6 +147,9 @@ def test_export_votes(db_session, tmp_path):
         ],
         responsible_committees=[Committee["AFET"]],
         result=VoteResult.ADOPTED,
+        amendment_subject=None,
+        amendment_number=None,
+        texts_adopted_reference="A10-123-123",
```
tillprochaska: Not relevant from a technical perspective as we don’t do anything with it, but I think this would be the expected format

```suggestion
        texts_adopted_reference="P10_TA(2025)0123",
```

backend/howtheyvote/export/__init__.py
```diff
@@ -382,6 +420,13 @@ def __init__(self, outdir: pathlib.Path):
             primary_key=["vote_id", "committee_code"],
         )
 
+        self.amendment_authors = Table(
+            row_type=AmendmentAuthorVoteRow,
+            outdir=self.outdir,
+            name="amendment_authors_votes",
+            primary_key=["vote_id"],
```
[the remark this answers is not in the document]
linusha: I misunderstood what these are - you are of course right.

Issue #1241, opened 2025-10-24
Amendment information (mostly) missing since 2025-10-09


linusha:
Number, Author, Subject are empty in the DB for nearly all amendments since above mentioned date. I think this is related to the VOTPipeline. Other fields are missing as well.

PR #1249, opened 2025-10-28
Collapse specific delimiters directly following each other

linusha:
This is the reason the VOT Pipeline for 22.10. was not scraped correctly. An amendment author tag contains a comma, directly followed by a newline, which breaks the parser. I could not visibly see this issue in my browser, but needed to download the file (and its visible in a debugger session ofc, but I wanted to check in the original). [Source](https://www.europarl.europa.eu/doceo/document/PV-10-2025-10-22-VOT_EN.xml)

I thought a bit about introducing this combination as a new delimiter, but I expect this to be a typo rather than something that happens regularly and I think the current solution has no real downside?

---

Some thoughts as a consequence of this:

- [x] After we merged this, I will rescrape the VOT list for this day. That will complete #1241. (I will look into why an earlier day failed, but that was for an unrelated reason, that much I know)
- [x] This pipeline failed with a ScrapingError after data was available. Again, getting a pushover notification for this would have been nice. I think, with the change we did on Sunday, this should happen now. Just checking in if you'd agree @tillprochaska, otherwise we should keep tabs on this.

### DIFF
--- a/backend/howtheyvote/scrapers/helpers.py
+++ b/backend/howtheyvote/scrapers/helpers.py
@@ -209,14 +209,18 @@ def parse_amendment_authors(raw_authors: str) -> list[AmendmentAuthor]:
     # If a newline or a comma is used as the delimiter, we can simply split the text
     if delimiter:
         for raw_author in raw_authors.split(delimiter):
-            authors.extend(parse_amendment_authors(raw_author))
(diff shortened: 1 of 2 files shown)


tillprochaska (commented):
Nice, thanks for debugging this! Wondering whether there will ever be an end to all the edge cases and typos 🙃

PR #1251, opened 2025-10-29
Generate share-picture for all votes

### LINKED ISSUE #1229: Enable OG image previews for non-main votes
I honestly am not sure why we disabled this until now. The obvious reason is that there is no way to discover these votes, but that will soon change with #1215. Are there technical reasons (time to generate, storage constraints) that made us not enable this?

When we change this, we should also brush over the alt text of the preview image. Right now, this will read the same for all amendments and the main vote, as only `display_title` is used in the text to describe the concrete vote.

linusha:
Adds the description into the alt text as well and generates share-pictures for all votes. That also means that we generate slightly more pictures than necessary (think of amendments where we dont currently recognize that they are amendments or the agenda motions). However, I think the number is so small in comparison, that it does not matter.

I generated pictures for 22.10 on my local machine and that resulted in close to 200 pictures, totalling 12.5MB.
Rough math yields 1.5GB for pictures for all votes we currently have in the DB.

Closes #1229.

### DIFF
--- a/backend/tests/worker/test_init.py
+++ b/backend/tests/worker/test_init.py
@@ -14,6 +14,10 @@ from howtheyvote.worker import get_worker
 
 
 def test_rcv_list_pipeline(db_session, mocker):
+    # Same as below, not patching these make the test extremely slow
+    mocker.patch("howtheyvote.pipelines.rcv_list.generate_vote_sharepics", return_value=None)
+    mocker.patch("howtheyvote.pipelines.vot_list.generate_vote_sharepics", return_value=None)
+
     plenary_session = PlenarySession(
         id="2025-07-07",
         start_date=datetime.date(2025, 7, 7),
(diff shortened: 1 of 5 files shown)


backend/tests/worker/test_init.py
```diff
@@ -14,6 +14,10 @@
 
 
 def test_rcv_list_pipeline(db_session, mocker):
+    # Same as below, not patching these make the test extremely slow
+    mocker.patch("howtheyvote.pipelines.rcv_list.generate_vote_sharepics", return_value=None)
+    mocker.patch("howtheyvote.pipelines.vot_list.generate_vote_sharepics", return_value=None)
+
```
tillprochaska: I couldn’t reproduce this, maybe we can look into this together tomorrow

tillprochaska (approved):
Thanks and sorry for the late review! I checked and apparently a few tens of thousands of files in a single directory should be fine on modern file systems. I ran this for all votes on my computer and it took ~70 minutes, and the size is approx. 1.43 GB as estimated by you.

PR #1276, opened 2025-11-15
Exchange accents from titles that are used as apostrophes

### LINKED ISSUE #1149: Prevent some line-breaks in Titles of Votes
Just a tiny thing: We should prevent line-breaks such as the one before 's. Link: https://howtheyvote.eu/votes/165965

Image

linusha:
Closes #1149.

After merging this, three votes need to be rescraped for this to become effective:

- [x] VOT + RCV 165965
- [x] RCV 155030
- [x] RCV 150777

### DIFF
--- a/backend/tests/scrapers/test_votes.py
+++ b/backend/tests/scrapers/test_votes.py
@@ -531,6 +531,24 @@ def test_procedure_scraper_multiple_responsible_committees(responses):
     assert fragment.data["responsible_committees"] == {"AFET", "BUDG"}
 
 
+def test_procedure_scraper_apostrophization(responses):
+    responses.get(
+        "https://oeil.secure.europarl.europa.eu/oeil/en/procedure-file?reference=2022/0377(NLE)",
+        body=load_fixture("scrapers/data/votes/oeil-procedure-file_2022_0377-nle.html"),
+    )
+
+    scraper = ProcedureScraper(
+        vote_id=155030, procedure_reference="2022/0377(NLE)", reference=None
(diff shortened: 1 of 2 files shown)


tillprochaska (approved):
Not super important, but it would be nice to have a test case for one of the scrapers covering this. For example, save this [procedure file](https://oeil.secure.europarl.europa.eu/oeil/en/procedure-file?reference=2022/0377(NLE)) as a fixture and assert that running the procedure scraper returns the expected procedure title.

In the past, errors like this were sometimes fixed in the source data, and when stumbling upon code handling edge cases like this, I’d sometimes wonder what the reason for implementing it was. I think a test case using the source data is a nice way of documenting that (and much faster than digging through web archives etc.)

backend/tests/scrapers/test_votes.py
```diff
@@ -531,6 +531,25 @@ def test_procedure_scraper_multiple_responsible_committees(responses):
...
+    responses.get(
+        "https://oeil.secure.europarl.europa.eu/oeil/en/procedure-file?reference=2022/0377(NLE)",
+        body=load_fixture("scrapers/data/votes/oeil-procedure-file_2022_0377-nle.html"),
+    )
+
+    scraper = ProcedureScraper(
+        vote_id=155030, procedure_reference="2022/0377(NLE)", reference=None
+    )
+    fragment = scraper.run()
+    print(fragment.data["procedure_title"])
```
tillprochaska: ☝️

Issue #1036, opened 2024-09-20
RSS support


kuchta:
It would be invaluable to have RSS support implemented, so that we can subscribe to it in our RSS readers.

Thank you for your much needed work ❤️

PR #1278, opened 2025-11-16
Provide latest 200 main votes as ATOM feed

linusha:
Generates feed (either by CLI or after scraping either new RCV or VOT lists) of he latest 200 main votes.
I chose 200 completely arbitrary, as I dont think a running feed of all votes makes that much sense and this should cover multiple months.

Each entry links to the /votes/id page and has text content reading like this:

> On Thursday, October 23 2025 Parliament voted on "Soil Monitoring and Resilience (Soil Monitoring Law)" (Proposition de rejet - Am 1= 2= 3=). There were 220 votes in favour, 341 votes in against, and 10 abstentions. The vote was REJECTED.

The last sentence is omitted if vote.result is not yet present (before VOT is available).

This could be much nicer (i.e. including summary text in the feed entry, partially rendering the page in the entries,...). However, I think this implementation provides the main value that people need right now, i.e., being notified when there are new votes, easily glance over them, see if there is something that interests them. I'd not spent too much energy improving this, unless we get specific requests.

I know that there is #1036, it's not linked on purpose - when we merged this, I'll write them a short message.

We should probably mention the feed somewhere, either in the footer or on /about or both... 
Do you have an opinion/idea @tillprochaska?


backend/howtheyvote/feed.py
```diff
@@ -0,0 +1,56 @@
...
+    def create() -> None:
+        query = (
+            select(Vote)
+            .where(Vote.is_main.is_(True))
+            .order_by(Vote.timestamp.desc())
+            .limit(200)
+        )
+        votes = Session.execute(query).scalars()
+
+        fg = FeedGenerator()
```
tillprochaska: Had a look at the package and found this: https://github.com/lkiesow/python-feedgen/issues/147

Not sure it is a problem now, but did you consider creating the feed yourself? As far as I know, Atom/RSS feeds are relatively simple XML documents, so it might be similarly easy to create them using the [standard lib](https://docs.python.org/3/library/xml.etree.elementtree.html#building-xml-documents) or Flask’s Jinja templates?

linusha: Yes, when I chose to use this I was aware that this is no longer maintained. I see that there are drawbacks to this, but given that Atom has been around a while, in the end I decided that's okay as the stuff we need works. 

However, I see the point that there are advantages from generating this inside of the flask app (which is what you meant with dynamic generation above, right?). I stumbled upon this rather late in the process and previously decided to not fiddle with a template myself as that's more complicated (even though the format is rather simple, I agree with you on that). Let me take another look at the necessary templating, then I'll come back with more thoughts. 

For now I guess the decision would be between shipping a first version using this lib or rewriting this to be generated in the flask app?

linusha:
@kuchta as someone who would like to have this feature, do you have an opinion on single votes as items versus one item per day of votes? 

I have to admit that making a single item per session did not occur to me previously, @tillprochaska. I tend to agree with your proposal though!

kuchta:
@linusha, since there's no page per session, I think it makes more sense to have one feed entry per vote, as each entry should contain a valid link. I understand that there can be dozen of votes in a session, but it is not uncommon for news sites to do exactly the same thing (generating dozens of entries in an instant) and it is very convenient to see subjects without having to look inside. If I just see a generic title like “Voting Results: Thursday, November 13, 2025,” my brain will likely skip over such an entry because it’s used to going through thousands of entries every day

backend/howtheyvote/feed.py
```diff
@@ -0,0 +1,56 @@
...
+            " represents just over 440 million Europeans. Although the Parliament publishes"  # noqa: E501
+            " information such as agendas, minutes, and vote results on its website, it can be"  # noqa: E501
+            " quite difficult to find out what MEPs voted on or how a particular vote turned out."  # noqa: E501
+            " HowTheyVote.eu compiles voting data from various official sources and allows anyone "  # noqa: E501
+            "to search for votes and view the results. This feed provides the 200 newest votes in our database."  # noqa: E501
+        )
+
+        for vote in votes:
+            fe = fg.add_entry(order="append")
+            fe.id("https://howtheyvote.eu/votes/" + str(vote.id))
```
tillprochaska:
```suggestion
            fe.id(frontend_url(f"/votes/{vote.id}"))
```

backend/howtheyvote/feed.py
```diff
@@ -0,0 +1,56 @@
...
+            .where(Vote.is_main.is_(True))
+            .order_by(Vote.timestamp.desc())
+            .limit(200)
+        )
+        votes = Session.execute(query).scalars()
+
+        fg = FeedGenerator()
+        # General properties of the feed
+        fg.id("http://howtheyvote.eu/")
+        fg.icon("http://howtheyvote.eu/files/feed_icon.png")
```
tillprochaska: This doesn’t seem to exist?

backend/howtheyvote/feed.py
```diff
@@ -0,0 +1,56 @@
...
+        for vote in votes:
+            fe = fg.add_entry(order="append")
+            fe.id("https://howtheyvote.eu/votes/" + str(vote.id))
+            fe.title(vote.display_title)
+            fe.pubDate(vote.timestamp.replace(tzinfo=ZoneInfo("Europe/Brussels")))
+            fe.link(href="https://howtheyvote.eu/votes/" + str(vote.id))
+            position_counts = count_vote_positions(vote.member_votes)
+            content = (
+                f'On {vote.date.strftime("%A, %B %d %Y")} Parliament voted on "'
+                f'{vote.display_title}" ({vote.description}). There were {position_counts["FOR"]}'  # noqa: E501
```
tillprochaska: I’d suggest to avoid "There were…" and instead use something like this:

> 220 MEPs voted in favor, 341 against, and 10 abstained. The vote was rejected.

or

> The vote was rejected with 220 votes in favor, 341 against, and 10 abstentions.

I guess technically, it’s not the vote that is rejected/adopted, but the report/amendment/resolution/…, but we don’t really know what it is in all case, so it’s probably easier to just keep the current wording.

backend/howtheyvote/api/votes_api.py
```diff
@@ -225,6 +227,65 @@ def index() -> ResponseReturnValue:
...
+    return {
+        "id": id,
+        "title": title,
+        "content": content,
+        "link": id,
+        "published": pub_date,
+    }
+
+
+@bp.route("votes/feed")
```
tillprochaska:
```suggestion
@bp.route("votes.atom")
```

or

```suggestion
@bp.route("votes/feed.xml")
```

backend/howtheyvote/api/votes_api.py
```diff
@@ -225,6 +227,65 @@ def index() -> ResponseReturnValue:
...
+def votes_feed() -> ResponseReturnValue:
+    """
+    ---
+    get:
+        operationId: getVotesFeed
+        summary: Get Votes as ATOM Feed
```
tillprochaska:
```suggestion
        summary: Get Votes as Atom Feed
```

PR #1279, opened 2025-11-17
Fix filter options without results

tillprochaska:
This fixes the issue described in this comment: https://github.com/HowTheyVote/howtheyvote/pull/1224#pullrequestreview-3327530785

For example, if a user selects the filters `country: Russia` and `committee: CONT` and there are no votes that match both filters, we will still include these options in the list because they are selected. This is what most search engines that provide faceted search seem to do in this case, and I think it makes sense.

As discussed before, the ideal solution would be to automatically refresh the available options whenever the selection for any of the other filters changes. This would prevent the situation in the first place, at least in most cases, and I still want to implement that as an additional enhancement on top of this fix.

### DIFF
--- a/backend/howtheyvote/api/query.py
+++ b/backend/howtheyvote/api/query.py
@@ -3,7 +3,7 @@ import datetime
 import enum
 from abc import ABC, abstractmethod
 from collections import defaultdict
-from typing import Any, Literal, Self, TypedDict
+from typing import Any, Literal, Self, TypedDict, overload
 
 from sqlalchemy import and_, asc, desc, func, select, true
 from sqlalchemy.sql import ColumnElement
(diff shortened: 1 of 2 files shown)


tillprochaska:
> this https://github.com/HowTheyVote/howtheyvote/pull/1283? No, right?

No, in that PR, I’m implementing something like this:

Screenshot 2025-11-29 at 6 42 53 PM

Whereas what I was referring to is automatically updating the options of facets in response to a user selecting one or more options in another facet:

https://github.com/user-attachments/assets/67962832-f005-4996-8a62-bd2ab68e3ec7

(After selecting "Russia", the "Categories" options are updated to only include categories that also match "Russia".)

PR #1281, opened 2025-11-19
Fix scheduled scraper tests

linusha:
The failing session scraper seems to just take a really long time sometimes. I tried it locally multiple times and it took me up to 80s to receive a response. As this is run seldomly, I just liberally increased the value a bit further.

Note that the scraper tests run against the live data sources fail even with this change. The reason for this is that https://oeil.secure.europarl.europa.eu/ redirects to https://oeil.europarl.europa.eu/ via a 307 response. I am not sure how temporary we think this is. 
Should I just go ahead and exchange (all?) of these URL values? Or only the ones in the tests? Or assert against either value? :thinking:


linusha:
> This is the same issue as https://github.com/HowTheyVote/howtheyvote/issues/1282, probably this is slowly being rolled out across all websites.

Can you check whether it works on your machine? It did on mine, with the flag set to not use the mocks.

> However, maybe we can wait a little longer whether we receive a reply from the EP admin before we spend time on this?

@tillprochaska They answered and won't provide actual solutions.

PR #1299, opened 2025-12-04
MEP profiles

tillprochaska:
**Merge #1283 first. Then update the target branch of this PR to `main`!**

This PR adds MEP profile pages. Profile pages list all votes an MEP participated in (where "participate" includes not voting) along with their vote position. Supports the same search, sorting, and filtering options as the main search page.

I’ve tried to reuse what we already have as much as possible, both on the backend and frontend. This also means that it should be much easier to add other "pre-filtered" listings of votes in the future (e.g., topic pages).

| Desktop | Mobile |
| --- | --- |
| Screen Shot 2025-12-04 at 21 54 17 | Screen Shot 2025-12-04 at 21 54 06 |

### DIFF
--- a/frontend/src/components/CopyrightLink.tsx
+++ b/frontend/src/components/CopyrightLink.tsx
@@ -1,0 +1,29 @@
+type CopyrightLinkProps = {
+  content: string;
+  sourceUrl: string;
+} & (
+  | { startYear: number; endYear: number; year?: never }
+  | { startYear?: never; endYear?: never; year: number }
+);
+
+function CopyrightLink({
+  content,
+  sourceUrl,
+  startYear,
+  endYear,
+  year,
+}: CopyrightLinkProps) {
+  return (
+    <>
+      {content}
+      {": © European Union "}
+      {year ? year : `${startYear}–${endYear}`}
+      {" · "}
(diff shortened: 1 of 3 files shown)


linusha:
I will look at this more closely of course, but is there a specific reason you opted to not use the contact information in the header, as with the original version? I don't have a set opinion yet, but wanted to ask first.

tillprochaska:
> but is there a specific reason you opted to not use the contact information in the header, as with the original version? I don't have a set opinion yet, but wanted to ask first.

No strong reasons. Primary reason is that it’s not a critical feature and I didn’t want to add more stuff to this PR as it already is quite big. Secondary reason is that I wasn’t sure whether we still want to push Facebook, Twitter, etc. Alternatively, we could for example reconcile with Wikidata to retrieve Bluesky and Mastodon handles.

frontend/src/components/MemberHeader.css
```diff
@@ -0,0 +1,74 @@
...
+  gap: var(--space);
+}
+
+.member-header__photo {
+  width: 33%;
+  max-width: 100px;
+  flex-grow: 0;
+  flex-shrink: 0;
+  align-self: center;
+  aspect-ratio: 4 / 5;
```
tillprochaska: Add `object-fit: cover`

PR #1326, opened 2025-12-09
Topics

tillprochaska:
This adds a model + boilerplate required for topics.

I’ve also added all topics (incl. subtopics) from https://www.europarl.europa.eu/topics/en/all to a data file. I’ve done that manually, because the list is relatively short, unlikely to change very frequently, and I can imagine that we’d deviate from the "original" over time anyway. While I’ve added all topics to the file, I’d actually suggest that we only consider the following:

* Climate and environment
* Animal welfare
* Biodiversity
* Circular economy
* Climate change
* Consumer rights/Consumer protection (not sure why one is a subtopic of the other and what exactly the difference is)
* Travel
* Digital (this is maybe a little generic?)
* Artificial Intelligence
* Online safety
* Disinformation
* Economy and budget
* EU budget
* Taxation
* Education
* Education, youth and culture
* Energy
* Food and agriculture
* Foreign affairs
* Enlargement
* International trade
* Gender equality
* Health
* Covid-19
* Medicines
* Migration
* Social protection
* Worker’s rights
* Terrorism and security
* Border controls
* Fighting terrorism

The remaining topics don’t make that much sense for our use case I think. Probably we could even eliminate a few more topics from the list above.

As mentioned before, for the mapping, I think it would even make sense to start with just a small subset of the topics in the list above, deploy that, and then iterate on it.

### DIFF
--- a/backend/howtheyvote/analysis/votes.py
+++ b/backend/howtheyvote/analysis/votes.py
@@ -5,7 +5,7 @@ from collections.abc import Iterable, Iterator
 from unidecode import unidecode
 
 from ..helpers import make_key
-from ..models import Fragment, PressRelease, Topic, Vote
+from ..models import Fragment, OEILSubject, PressRelease, Topic, Vote
 from ..vote_stats import count_vote_positions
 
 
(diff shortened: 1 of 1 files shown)


backend/howtheyvote/analysis/votes.py
```diff
@@ -247,3 +247,64 @@ def _match_by_position_counts(self, vote: Vote) -> set[str]:
...
+
+    def run(self) -> Fragment | None:
+        if self.vote.oeil_subjects == []:
+            return None
+
+        topics = []
+
+        oeil_subjects = " ".join([subject.code for subject in self.vote.oeil_subjects])
+
+        subject_to_topic = {
```
linusha: Long term it might make sense to split this up into different mappings for different topics, I find this feasible for the moment...

tillprochaska: I think it’s fine like this, though you could make the mapping a class variable of `TopicsAnalyzer` so that the dict isn’t recreated for every single instance.

tillprochaska:
I left a few comments with implementation suggestions, but I agree that we shouldn’t mind getting the mappings perfect before we deploy.

It would also be cool to exclude the topics in the export. (Although it might have been an explicit choice not to do that until the mappings are still a little experimental?)

backend/howtheyvote/analysis/votes.py
linusha: :ballot_box_with_check:

backend/howtheyvote/api/serializers.py
```diff
@@ -84,6 +85,21 @@ def serialize_country(country: Country) -> CountryDict:
...
 
+class TopicDict(TypedDict):
+    code: Annotated[str, "climate-change"]
+    """Unique identifier of the topic"""
+
+    label: Annotated[str, "Climate change"]
+    """Readable name of the topic"""
+
+    parent_code: Annotated[str | None, "climate-and-environment"]
+    """In the case of second-level topic, the code of the upper hierarchy."""
```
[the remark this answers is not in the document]
linusha: :ballot_box_with_check:

backend/howtheyvote/api/serializers.py
```diff
@@ -84,6 +85,21 @@ def serialize_country(country: Country) -> CountryDict:
     }
 
 
+class TopicDict(TypedDict):
+    code: Annotated[str, "climate-change"]
+    """Unique identifier of the topic"""
```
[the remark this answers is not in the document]
linusha: > Maybe we could add a note saying that the code might change, at least for now? Or alternatively a note on the topics property of the vote schema highlighting that we’re actively working on the classification, and users have to expect topics and how they are assigned to change.

Yes, I will do that.

backend/howtheyvote/analysis/votes.py
```diff
@@ -247,3 +247,64 @@ def _match_by_position_counts(self, vote: Vote) -> set[str]:
...
+    def __init__(self, vote: Vote):
+        self.vote = vote
+
+    def run(self) -> Fragment | None:
+        if self.vote.oeil_subjects == []:
+            return None
+
+        topics = []
+
+        oeil_subjects = " ".join([subject.code for subject in self.vote.oeil_subjects])
```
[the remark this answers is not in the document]
linusha: I find this to be much more complex, but I'll change it, this is definitely more robust.

backend/howtheyvote/analysis/votes.py
```diff
@@ -247,3 +247,74 @@ def _match_by_position_counts(self, vote: Vote) -> set[str]:
...
+
+        oeil_subjects = set()
+        for subject in self.vote.oeil_subjects:
+            oeil_subjects.add(subject)
+            oeil_subjects.update(subject.parents)
+
+        topics = []
+        for subject in oeil_subjects:
+            if subject.code in TopicsAnalyzer.subject_to_topic:
+                topics.append(TopicsAnalyzer.subject_to_topic[subject.code])
```
linusha: @tillprochaska not sure if I understood you correctly, but this `.code` was necessary the way it is implemented right now.

tillprochaska: Yes, that’s right! You could also use `OEILSubject` objects as the dict keys in which case it can be omitted here. This has the benefit that any invalid subject code would cause immediate failure, but I think it’s also fine this way.

Issue #1334, opened 2025-12-16
Sharepic generation upon scraping failed


linusha:
As per the notification send out today. I just went on the server and created them manually, with a quick edit to the CLI command. The problem is thus not with the chromium container but something worker related probably.

PR #1337, opened 2025-12-17
Fix sharepic generation in worker

tillprochaska:
You can reproduce the error (and verify the fix) by running the RCV pipeline inside of the `worker` container:

```
docker compose exec worker bash
htv pipeline rcv-list --term=10 --date=2025-11-13
```

Fixes https://github.com/HowTheyVote/howtheyvote/issues/1334


linusha (approved):
Sorry, I could have caught that during my review of the initial change.

PR #1339, opened 2025-12-18
Remove non-breaking spaces from vote titles

### LINKED ISSUE #1336: Formatting breaks for one VoteCard title
image

linusha:
Closes #1336.

image

Here's how it looks locally after scraping today.

I'd like to also add a test for the scraper with todays XML file, but working with that file is extremely painful.
I'll do that though, would appreciate merging this earlier if you are alright with the change and add the test potentially later. Looks quite bad on the site right now on mobile devices. 

I missed the failing test, my proposal would be to change that to also not expect a non-breaking space. I see that in that instance, non-breaking makes sense, but I don't think there is a way to reliably sort these cases apart from the ones were they were just used by mistake? :thinking:

### DIFF
--- a/backend/tests/scrapers/data/votes/oeil-procedure-file_2025_0229-nle.html
+++ b/backend/tests/scrapers/data/votes/oeil-procedure-file_2025_0229-nle.html
@@ -1,0 +1,2359 @@
+<!DOCTYPE html>
+<html lang="en" data-lt-installed="true"><head>
+
+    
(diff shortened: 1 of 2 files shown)


tillprochaska:
> I'd like to also add a test for the scraper with todays XML file, but working with that file is extremely painful.

Yes, sounds good

tillprochaska:
> I missed the failing test, my proposal would be to change that to also not expect a non-breaking space.

👍

tillprochaska (approved):

PR #1348, opened 2026-01-02
Preaggregated CSVs for Countries and Groups

### LINKED ISSUE #1128: Provide pre-aggregated CSV downloads
We currently provide the a CSV file with the voting behavior of individual MEPs, for example: https://howtheyvote.eu/api/votes/172663.csv

In addition, it would be nice to provide CSV downloads with the aggregated voting behavior by political group and country. This would make it easier to e.g. visualize vote results for a single vote in tools like Datawrapper.

linusha:
Closes #1128.

In the end I just declared the old and new routes for the members CSV. I initially wanted to use a redirect, that I got rid of that was mainly due to laziness, we could do it if you'd like. My initial reasoning was that we ideally would want people to use the `/members.csv` endpoint in case we would eventually get rid of the old one.

I declared the content-disposition headers to specify the file-names for the downloads, as otherwise the default filename in my browser would have just been "members.csv", etc. **What we should somehow test (or possibly you have a guess) is whether this breaks the datawrapper integration?** In that case we need to find another solution for the filename issue.

Also, I have not yet added a mention of this in the frontend. I am also not too sure whether I'd do that...

### DIFF
--- a/frontend/src/lib/links.ts
+++ b/frontend/src/lib/links.ts
@@ -30,7 +30,7 @@ export function getDownloadUrl(voteId: number, format: DataFormat): string {
   const url = new URL(`/api/votes/${voteId}`, BACKEND_PUBLIC_URL);
 
   if (format === "csv") {
-    return `${url.toString()}.csv`;
+    return `${url.toString()}/members.csv`;
   }
 
   return url.toString();


linusha:
> I’m fine with this. I think the "/members.csv" route would be the default route that’s displayed in the API docs as implemented.

Yup.
image

> Maybe you could update the link on vote pages in the "Open data" section as well?

Done!

PR #1356, opened 2026-01-08
Match OEIL summaries and votes based on position counts

### LINKED ISSUE #1307: Handle ambiguous OEIL summaries
Another thing we might want to consider in a future iteration:

It is possible that there are two votes related to the same procedure on the same day. In that case, there may be a single OEIL summary for both votes, or there may be two separate summaries. I think this PR already handles the latter case.

Example for 2 votes/1 summary:

| Vote 1 | Vote 2 | Summary |
| --- | --- | --- |
| https://howtheyvote.eu/votes/142140 | https://howtheyvote.eu/votes/142811 | https://oeil.europarl.europa.eu/oeil/en/document-summary?id=1703305 |
| https://howtheyvote.eu/votes/168456 | https://howtheyvote.eu/votes/168264#more-information | https://oeil.europarl.europa.eu/oeil/en/document-summary?id=1784597 |
| https://howtheyvote.eu/votes/164815 | https://howtheyvote.eu/votes/165216 |https://oeil.europarl.europa.eu/oeil/en/document-summary?id=1777377 |

```sql
SELECT v1.procedure_reference, DATE(v1.timestamp), Date(v2.timestamp), v1.id, v2.id
FROM votes v1
JOIN votes v2 ON (
	v1.id != v2.id AND
	v1.is_main = true AND
	v2.is_main = true AND
	v1.procedure_reference = v2.procedure_reference AND
	DATE(v1.timestamp) = DATE(v2.timestamp)
)
```

_Originally posted by @tillprochaska in https://github.com/HowTheyVote/howtheyvote/issues/1266#issuecomment-3620516605_

tillprochaska:
A OEIL summary is now matched with a vote if the following conditions are met:

* The summary mentions the result of exactly one RCV (for/against/abstentions), and this matches the result of the vote.
* The procedure reference and date of the summary and the vote match.

This handles the following edge cases:

* There are two RCV votes on the same day related to the same procedure (example: [vote 1](https://howtheyvote.eu/votes/149084), [vote 2](https://howtheyvote.eu/votes/148972)). In this case, the summary is matched to neither of them.
* There are two RCV votes on the same day related to the same procedure, but the summary is only about one of them (example: [vote 1](https://howtheyvote.eu/votes/182713), [vote 2](https://howtheyvote.eu/votes/182767)). In this case the summary is matched to the correct vote.

I found a small number of OEIL summaries (~5 per year) that are about an RCV vote but do not mention the exact position counts. These summaries wouldn’t be matched anymore. However, for most of them, a press release was available as an alternative. To handle this, we could consider also matching summaries and votes if a summary doesn’t mention any position counts at all *and* there is only one RCV vote for the same date and procedure. Given the small number of summaries this affects and to keep things simple, I haven’t done that for now.

Fixes #1307

### DIFF
--- a/backend/howtheyvote/analysis/helpers.py
+++ b/backend/howtheyvote/analysis/helpers.py
@@ -30,6 +30,7 @@ def parse_int(text: str) -> int:
     return int(text)
 
 
+# Example: The text was adopted by 400 votes in favour, 63 against with 81 abstentions.
 VOTE_RESULT_REGEX_1 = re.compile(
     NUMBER_REGEX
     + r"\s(?:votes?|MEPs? voted)\s(?:in\sfavou?r|for),\s"
@@ -40,6 +41,8 @@ VOTE_RESULT_REGEX_1 = re.compile(
     flags=re.I,
 )
 
+# Example: The European Parliament decided by 345 votes to 284, with 8 abstentions, to
+# refuse to grant discharge…
 VOTE_RESULT_REGEX_2 = re.compile(
     r"(?:with|by)\s"
     + NUMBER_REGEX


linusha:
> After deploying this, run the following:
> 
> ```
> htv flush votes --source-name=OEILSummaryIDScraper --source-name=OEILSummaryScraper
> htv temp oeil-summaries
> ```

@tillprochaska I don't think that suffices? :thinking: The pipeline for the OEIL summaries operate on the vote level, so just removing the fragments it not enough, no? Wouldn't we need to execute the `htv temp oeil-summaries` but with `force` set to `True`?

PR #1363, opened 2026-01-11
Datawrapper Embeds for Groups and Countries

linusha:
## Groups

image

## Countries

image

## Embed Section

image

---

1. I personally would have preferred to have the country chart rotated as I find that easier to read than a very long list, but the stacked column chart works very different than the stacked bar chart and I did not find a way to rotate that one.
2. Having flags in the chart would have also been cool, but that required having emoji codes for the flags in the data. Maybe some other time...
3. The code for the `DataWrapperLinkList` is for sure not the cleanest solution to this. I would appreciate an opinion on what you say would be the best implementation.

### DIFF
--- a/frontend/src/client.entry.ts
+++ b/frontend/src/client.entry.ts
@@ -1,6 +1,6 @@
 import Banner from "./components/Banner";
 import CountryStatsList from "./components/CountryStatsList";
-import DatawrapperLink from "./components/DatawrapperLink";
+import DatawrapperLink from "./components/DatawrapperLinkList";
 import Eyes from "./components/Eyes";
 import GroupStatsList from "./components/GroupStatsList";
 import MemberVotesList from "./components/MemberVotesList";
--- a/frontend/src/components/DatawrapperLinkList.tsx
+++ b/frontend/src/components/DatawrapperLinkList.tsx
@@ -1,10 +1,9 @@
 import type { ComponentChildren } from "preact";
 import type { MouseEventHandler } from "preact/compat";
 import type { Vote, VotePositionCounts } from "../api";
-import List from "../components/List";
-import ListItem from "../components/ListItem";
 import { formatDate } from "../lib/dates";
-import { Island } from "../lib/islands";
+import List from "./List";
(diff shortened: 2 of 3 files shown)


frontend/src/components/DatawrapperLink.tsx
```diff
@@ -61,47 +69,157 @@ function DatawrapperLink({
...
+              reference={vote.reference}
+              description={vote.description}
+              stats={vote.stats.total}
+            >
+              Votes by groups
+            </DatawrapperLink>
+          </Island>
+        }
+        subtitle={
+          "An interactive barchart showing how members of each group voted."
```
tillprochaska: Wording suggestion:

```suggestion
          "A bar chart showing how MEPs of each political group voted."
```

I think we use "MEP" instead of "member" in most places in the frontend, and "political group" instead of just "group". I’d also remove "interactive". I think interactive visualizations are typically associated with being able to search/filter/change parameters of the visualization rather than just showing a tooltip/hover effects.

frontend/src/components/DatawrapperLink.tsx
```diff
@@ -61,47 +69,157 @@ function DatawrapperLink({
...
+              reference={vote.reference}
+              description={vote.description}
+              stats={vote.stats.total}
+            >
+              Votes by countries
+            </DatawrapperLink>
+          </Island>
+        }
+        subtitle={
+          "An interactive barchart showing how MEPs from the different member states voted."
```
tillprochaska:
```suggestion
          "A bar chart showing how MEPs from different countries voted."
```

(We use "countries" instead of "member states" elsewhere.)

tillprochaska:
> I personally would have preferred to have the country chart rotated as I find that easier to read than a very long list, but the stacked column chart works very different than the stacked bar chart and I did not find a way to rotate that one.

Another argument for a bar chart is that it works much better on narrow/mobile screens.

> The code for the DataWrapperLinkList is for sure not the cleanest solution to this. I would appreciate an opinion on what you say would be the best implementation.

I’ll take a look!

PR #1364, opened 2026-01-11
Fix missing amendments

tillprochaska:
I’m not entirely sure how this happened (see commit messages), but this should fix it:

```
htv temp vote-groups
htv aggregate votes
```


linusha (approved):
Honestly, I have not thought trough how we ended up in this situation.
But I can confirm that after running the commands you supplied in the original comment, amendments were accessible on my localhost page without any further scraping.

PR #1385, opened 2026-02-12
Make links in lists underlined by default

tillprochaska:
Currently, the underline becomes visible only when hovering over a link. I think I originally implemented that to reduce visual clutter. However, it also obscures the fact that they are actually clickable links (e.g., the names of MEPs or the titles in the "More information" section), especially given that most other links are underlined by default.

The links in the page nav still don’t have underlines by default. I think because it’s a navigation bar, it’s much clearer from context that these are links, but I wouldn’t mind making them underlined by default as well if you prefer

### DIFF
--- a/frontend/src/components/ListItem.css
+++ b/frontend/src/components/ListItem.css
@@ -24,13 +24,3 @@
 .list-item__chart {
   margin: var(--space-xxs) 0;
 }
-
-.list-item a {
-  text-decoration: none;
-}
-
-.list-item a:hover,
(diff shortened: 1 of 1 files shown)


linusha (approved):
For future reference, here's how the navbar would look like with always underlined links:

Image

I agree with you that hopefully its obvious enough that those are links, as I feel that actually looks quite cluttered this way. If we find that people think this is unintuitive, we can change it any minute, but for now I like the way you did it.

PR #1393, opened 2026-02-14
Fix feed MIME type

tillprochaska:
Not sure it makes a practical difference, but as this is an Atom feed, I think this would be the appropriate content type?


linusha (approved):
When researching for the implementation, my impression was that most people do Atom nowadays and call that RSS regardless, so I am not sure whether there are actual benefits. But I agree, this is technically correct and I don't think there is any harm in doing it.

PR #1394, opened 2026-02-14
Add link to feed in footer

tillprochaska:
Technically, it’s an Atom feed and not RSS, but I guess RSS is more commonly used?

### DIFF
--- a/frontend/src/components/Footer.tsx
+++ b/frontend/src/components/Footer.tsx
@@ -39,6 +39,15 @@ export default function Footer({ style, copyright }: FooterProps) {
           </a>
         </li>
         <li>
+          <a
+            href="/api/votes/feed.xml"
+            target="_blank"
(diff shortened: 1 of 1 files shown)


frontend/src/components/Footer.tsx
```diff
@@ -38,6 +38,15 @@ export default function Footer({ style, copyright }: FooterProps) {
             Bluesky
           </a>
         </li>
+        <li>
+          <a
+            href="/api/votes/feed.xml"
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            RSS
```
linusha: "Feed" would imo also be recognizable and dodges the RSS/Atom problem.

tillprochaska: Let’s keep RSS then, I think that’s clearer than just feed.

PR #1441, opened 2026-05-10
Handle members without a group membership

### LINKED ISSUE #1440: Handle members without group memberships
Most MEPs should have at least one group membership. (Even if they are non-attached members, they’d have a group membership with the `NI` group.) However, there are some members (?) for which the EP website doesn’t list a group membership.

Jeppe Kofod (https://www.europarl.europa.eu/meps/en/124870/JEPPE_KOFOD/history/9): If I understand correctly, he was reelected but left the EP almost immediately after that.

Frans Timmermans (https://www.europarl.europa.eu/meps/en/129141/FRANS_TIMMERMANS/history/9): Maybe because he was the lead candidate in 2019?

Right now, trying to access the member profile page results causes an unhandled error. I would suggest we treat these members as if they had never been a member and simply return a 404 from the API for any member without at least one group membership.

tillprochaska:
Closes #1440

### DIFF
--- a/backend/howtheyvote/api/members_api.py
+++ b/backend/howtheyvote/api/members_api.py
@@ -39,7 +39,7 @@ def show(member_id: int) -> ResponseReturnValue:
     member = Session.get(Member, member_id)
     today = datetime.date.today()
 
-    if not member:
+    if not member or not member.group_memberships:
         return abort(404)
 
     group = member.group_at(today) or member.group_memberships[-1].group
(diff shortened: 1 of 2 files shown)


linusha (approved):

linusha:
I assume you also did that, but just for documentation purposes: I checked, and those are the only two members that have no group-membership.

Both were elected, but did never act on their mandate.

linusha:
That Timmermans has a profile is a bit funny, [this Spanish MEP](https://www.europarl.europa.eu/meps/en/202352/ANTONI_COMIN+I+OLIVERES/history/9)  was technically also elected but does not have a profile for term 10. Huh.

PR #1451, opened 2026-05-21
Workaround TLS fingerprints

tillprochaska:
The WAF seems to use TLS fingerprinting techniques similar to [JA3/JA4](https://engineering.salesforce.com/tls-fingerprinting-with-ja3-and-ja3s-247362855967/) to detect requests from automated clients. To work around that, we try to send requests in a way that is as close as possible to a real browser (e.g., by using the same default headers and same TLS ciphers).

That alone isn’t enough. Usually we still get presented with a JS challenge. The JS challenge isn’t active (i.e., it doesn’t require active input like a captcha does). But it needs a real browser that can execute JavaScript. When the challenge is passed, a cookie is set that can be used to bypass the challenge on subsequent requests. So this PR also repurposes the headless Chromium instance we already use to generate sharepics to acquire the WAF cookie.

- [ ] Figure out what to do with scheduled tests. We could use the same method to acquire a WAF token. But I’d probably do that in a separate PR.
- [x] Refactor `get_url` in `scrapers.common`
- [x] Request sleep in tests
- [ ] ~Write tests~ (I would be fine with not testing the CDP classes. The tests would probably be rather complex, and we’ll notice when they break anyway…)


backend/howtheyvote/pipelines/rcv_list.py
```diff
@@ -59,11 +64,17 @@ def __init__(
...
+        self._ep_aws_waf_token = solve_ep_aws_waf_challenge()
+
+        # Temporarily disable EUR-Lex scrapers
+        # self._eurlex_aws_waf_token = solve_eurlex_aws_waf_challenge()
+        self._eurlex_aws_waf_token = None
+
         self._scrape_rcv_list()
         self._scrape_documents()
-        self._scrape_eurlex_documents()
+        # self._scrape_eurlex_documents()
```
tillprochaska, line 75: Commented out for now. These are removed entirely in #1472

backend/howtheyvote/pipelines/rcv_list.py
```diff
@@ -59,11 +64,17 @@ def __init__(
         self._request_cache: RequestCache = LRUCache(maxsize=25)
 
     def _run(self) -> None:
+        self._ep_aws_waf_token = solve_ep_aws_waf_challenge()
+
+        # Temporarily disable EUR-Lex scrapers
+        # self._eurlex_aws_waf_token = solve_eurlex_aws_waf_challenge()
+        self._eurlex_aws_waf_token = None
+
```
linusha, line 72: This can then go as well, when you combine the PRs :)

backend/howtheyvote/waf.py
```diff
@@ -0,0 +1,171 @@
...
+
+def solve_ep_aws_waf_challenge() -> str:
+    # Requests to /doceo/* seem to always trigger the JS challenge (in contrast to
+    # other pages such as the homepage).
+    return solve_aws_waf_challenge(
+        "https://www.europarl.europa.eu/doceo/document/A-10-2026-0011_EN.html"
+    )
+
+
+def solve_eurlex_aws_waf_challenge() -> str:
```
linusha, line 108: Do you want to keep this? I don't have an opinion.

backend/howtheyvote/pipelines/rcv_list.py
tillprochaska: Yes, just didn’t want to completely remove it in this PR as that would cause more merge conflicts when merging the other branch.

backend/howtheyvote/waf.py
tillprochaska: No

docker-compose.yml
```diff
@@ -2,21 +2,11 @@ volumes:
...
-  # This network is used to limit network access of the `chromium` service
-  # to just the `backend` and `frontend` services, disabling internet access.
-  chromium:
-    driver: bridge
-    internal: true
-
 services:
   frontend:
-    image: "ghcr.io/howtheyvote/howtheyvote-frontend:main"
+    image: "ghcr.io/howtheyvote/howtheyvote-frontend:${HTV_TAG:-main}"
```
linusha, line 7: I am a bit confused right now - I can see that this is supported in our playbook, but the commit in there is quite old already. Did we just never use this until now?

linusha (approved):

linusha:
Ah, forgot to write that:

> I would be fine with not testing the CDP classes. The tests would probably be rather complex, and we’ll notice when they break anyway…

Agreed.

 > Figure out what to do with scheduled tests. We could use the same method to acquire a WAF token. But I’d probably do that in a separate PR.
 
 Yes, for now I'll open a ticket.

docker-compose.yml
tillprochaska: Hm, not sure, I only added support for this to the playbook when I also made this change. And it refers to a branch name rather than a commit hash: https://github.com/HowTheyVote/infra/commit/ea9bbb5f3b926683a826810d2d802762449711d2

linusha: Sorry, ignore me. I simply forgot that this PR has been around for a while already. Mea culpa.

PR #1466, opened 2026-06-05
Group votes based on session start date (instead of vote date)

### LINKED ISSUE #1465: Relax criteria for grouping votes
Currently, we use some combination of (I think) title and date. That breaks if amendment votes and final votes are cast on different dates. We should probably relax that a little bit and only require that the votes where cast during the same plenary session.

tillprochaska:
Amendment votes and final votes may be cast on different days of the same plenary session. Currently, in such a case, the amendment votes and final vote wouldn’t be grouped together (i.e., no amendments would be displayed on the website). This changes the analyzer to group based on the plenary session start date. So if there are multiple votes with the same title that were all cast during the same plenary session, they will all be grouped together.

Fixes #1465

Run the following commands to recompute group keys:

```
htv temp vote-groups
htv aggregate votes
```

After that, amendments should be displayed for [this vote](http://localhost:8000/votes/131714), even though they were cast on different days.

Todos:

- [x] Order votes on amendments page by date

### DIFF
--- a/backend/tests/analysis/test_votes.py
+++ b/backend/tests/analysis/test_votes.py
@@ -41,7 +41,7 @@ def test_vote_group_analyzer_title():
     ]
 
     analyzer = VoteGroupsAnalyzer(
-        date=datetime.date(2026, 1, 1),
+        session_start_date=datetime.date(2026, 1, 1),
         votes=votes,
     )
     fragments = list(analyzer.run())
@@ -75,7 +75,7 @@ def test_vote_group_analyzer_dlv_title():
     ]
 
     analyzer = VoteGroupsAnalyzer(
-        date=datetime.date(2026, 1, 1),
+        session_start_date=datetime.date(2026, 1, 1),
         votes=votes,
     )
     fragments = list(analyzer.run())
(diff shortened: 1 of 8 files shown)


backend/tests/api/test_votes_api.py
```diff
@@ -1002,6 +1002,39 @@ def test_votes_api_related_votes(db_session, api):
     assert res.json["related"] == expected
 
 
+def test_votes_api_related_votes_order(db_session, api):
+    amendment = Vote(
+        id=130991,
+        is_main=False,
+        timestamp=datetime.datetime(2021, 5, 18, 0, 0, 0),
+        order=261,
```
linusha, line 1010: A question only semi-related to this PR: Could you remind me of where this comes from? Is that just the order in which the votes appear in the RCV list? 🤔 
I tried searching for `order:` and `order =` in the code but to no success...

backend/tests/analysis/test_votes.py
```diff
@@ -86,6 +86,36 @@ def test_vote_group_analyzer_dlv_title():
...
+
+    analyzer = VoteGroupsAnalyzer(
+        session_start_date=datetime.date(2021, 5, 17),
+        votes=votes,
+    )
+    fragments = list(analyzer.run())
+
+    assert len(fragments) == 2
+    assert fragments[0].group_key == 130991
+    assert fragments[1].group_key == 131714
```
linusha, line 115: And a second learning question, just to be safe: This group key tells me that the fragments are from two separate entities, is that correct? I.e., fragments with the same group__key are merged, these not - which makes sense as those are two different votes, and that they are related becomes relevant only in a subsequent step.

linusha (commented):
🚀 
Can confirm - there are amendments now :)

backend/tests/api/test_votes_api.py
tillprochaska: Yes, exactly, it is the order in the RCV list: https://github.com/HowTheyVote/howtheyvote/blob/main/backend/howtheyvote/scrapers/votes.py#L122

It should correspond to the order of timestamps for recent RCV data. But I think older RCV lists do not have exact timestamps of the votes, but only dates, so we needed a separate field to retain the order of the votes in the original source.

backend/tests/analysis/test_votes.py
tillprochaska: Ah yes, didn’t even think about this when writing the test, but I remember that we discussed this naming collision. But you’re understanding is exactly correct:

* `Fragment.group_key` is what defines which fragments contain data about the same record. In this case it’s `Vote` records, but every fragment for any record type has a group key.
* The `group_key` data field is specific to `Vote` records and is a hash of session date + title. This is what we use to group together votes related to the same agenda item (usually amendment votes and the final vote).

tillprochaska: I’m not entirely sure what the best way to resolve this collision might be. Maybe simply renaming `Fragment.group_key` to `record_id` or something similar?

linusha (approved):
Sorry, this should have been an approval in the first place, forgot to switch the option.

PR #1468, opened 2026-06-08
Add national parties

### LINKED ISSUE #1399: National parties
This list includes national political parties (ID, short and long labels, start and end dates): https://data.europarl.europa.eu/en/datasets/corporate-bodies-of-the-european-parliament/91

Files are versioned and I’m not sure if there’s a URL that always redirects to the latest version.

Alternatively, the following API can be used: https://data.europarl.europa.eu/api/v2/corporate-bodies?body-classification=NATIONAL_POLITICAL_GROUP&format=application%2Fld%2Bjson&offset=0&limit=1000

However, I’m not sure if it’s possible to include additional information in the list or if we’d have to send separate requests for that, so it might be easier to use the CSV table from above.

linusha:
# Still to do:

- [ ] integrate in members pipeline
- [ ] fix tests
- [ ] integrate into export


backend/howtheyvote/cli/dev.py
```diff
@@ -510,6 +510,59 @@ def load_oeil_subjects(file: TextIO) -> None:
     subjects.save()
 
 
+@dev.command()
+def load_national_parties() -> None:
+    # TODO: Update this
+    """Loads a list of national parties as published by the EP Open Data Portal.
+    """
```
tillprochaska:
```suggestion
    """Loads a list of national parties as published by the EP Open Data Portal."""
```

backend/howtheyvote/cli/dev.py
```diff
@@ -510,6 +510,59 @@ def load_oeil_subjects(file: TextIO) -> None:
...
+    parties = DataclassContainer(
+        dataclass=NationalParty,
+        file_path=DATA_DIR.joinpath("national_parties.json"),
+        key_attr="id",
+    )
+
+    meta_data = requests.get(
+        "https://data.europarl.europa.eu/OdpDatasetService/Datasets/corporate-bodies-of-the-european-parliament"
+    ).json()
```
tillprochaska: Just to fix the ugly formatting

```suggestion
    meta_url = "https://data.europarl.europa.eu/OdpDatasetService/Datasets/corporate-bodies-of-the-european-parliament"
    meta_data = requests.get(meta_url).json()
```

backend/howtheyvote/cli/dev.py
```diff
@@ -510,6 +510,59 @@ def load_oeil_subjects(file: TextIO) -> None:
...
+    latest_version = len(meta_data["odpDatasetVersions"])
+
+    data_url = f"https://data.europarl.europa.eu/distribution/corporate-bodies_{latest_version}_en.csv"
+
+    with requests.get(data_url, stream=True) as r:
+        lines = (line.decode("utf-8") for line in r.iter_lines())
+        # Structure of each line:
```
tillprochaska: Can you use a `csv.DictReader` instead? That would allow you to access the columns using the column names from the header (e.g., `row["corporate_body_label"]`). It’s more readable, and it will also break when the structure of the CSV changes.

backend/howtheyvote/models/national_parties.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+@dataclasses.dataclass(frozen=True)
+class NationalParty(DeserializableDataclass, metaclass=NationalPartyMeta):
+    id: str
+    short_label: str
+    label: str
+    alt_label: str | None
+    # TODO: Fix type here?
+    start_date: str | None
```
tillprochaska: Is there any case where a national party doesn’t have a start date?

linusha:
@tillprochaska Here are some screenshots of how it currently looks:

image
Screenshot from 2026-08-03 19-15-01

For the members pipeline, either #1535 eventually works (which it currently does not) or I am actively considering moving this to the ODP. The problematic part of this would be the group memberships, I need to think about this.

tillprochaska:
@linusha Nice, looks good! Do we have the abbreviations (e.g., SPD instead of Sozialdemokratische Partei Deutschlands)? If yes, we could show the abbreviation and the full name in a tooltip (similar to the group name).

linusha:
Yes we have the abbreviations, which is why you can search for them ;)

I will add the tooltip, great call!

Issue #1381, opened 2026-02-09
SPARQL Query to retrieve EuroVoc is estimated to be too expensive


linusha:
The last automated update of the EuroVoc file failed, see [here](https://github.com/HowTheyVote/howtheyvote/actions/runs/21554601843/job/62108755839). I can reproduce this locally and when examining the response, one can see that the query is estimated as too expensive in terms of execution time and is thus blocked.

I am either time-outed for a while or the whole endpoint is currently not stable, so I can't test further right now, but I think one workaround would be to split the query into smaller ones and then take care of the merging locally. I think I would start with retrieving the geo_area_code separately, and then joining that onto the rest. If that does not suffice, the same could be done with the related and broader ids. My theory would be that those especially are quite expensive (together with alt label), as we have the distinct concatenation in the query? Just spitballing here though.

If you don't intervene @tillprochaska, I will pursue this further, trying to reduce the estimated cost of individual queries and managing the rest on our side...

tillprochaska:
I just tried the query without the geo area alignment directly on [data.europa.eu/sparql](https://data.europa.eu/data/sparql?locale=en), and I still get the error. Manually increasing the execution timeout doesn’t help. Given that this query has usually completed in ~5s, the estimate seems to be off by a lot. I’d maybe try filing a bug with the Publications Office helpdesk first?

While we could of course then send a separate request to fetch related/broader concepts, etc., I doubt that would be less expensive in total, and it also completely defeats the purpose of SPARQL.

```
Virtuoso 42000 Error The estimated execution time 851 (sec) exceeds the limit of 120 (sec).

SPARQL query:
    PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
    PREFIX dc: <http://purl.org/dc/elements/1.1/>
    PREFIX euvoc: <http://publications.europa.eu/ontology/euvoc#>

    SELECT
      ?term
      ?label
      (GROUP_CONCAT(DISTINCT ?related_id, ",") as ?related_ids)
      (GROUP_CONCAT(DISTINCT ?broader_id, ",") as ?broader_ids)
      (GROUP_CONCAT(DISTINCT ?alt_label, ",") as ?alt_labels)
    WHERE {
      GRAPH <http://publications.europa.eu/resource/dataset/eurovoc> {
        ?term euvoc:status <http://publications.europa.eu/resource/authority/concept-status/CURRENT>.

        ?term skos:prefLabel ?label_.
[...]

PR #1480, opened 2026-06-22
Reduce complexity of Eurovoc SPARQL query

tillprochaska:
Removing the `GRAPH` patterns seems to reduce the (estimated) cost enough to bring it below the server limit for `PUBLICATIONS_ENDPOINT`. (It does get around the limit for `DATA_ENDPOINT` as well, but the endpoint then times out executing the query.) As both endpoints can provide the data we need, I’ve switched to `PUBLICATIONS_ENDPOINT`.

As the query now matches against the entire dataset and not just the specified graphs, I had to slightly adjust it. For example, I’ve added new triples to fix the `rdf:type` of `?term` and `?geo_area` ensuring the result stays the same as before.


linusha (approved):
This is really cool, thanks a lot! 🙏 

Imo, this could close #1381 then?

tillprochaska:
Indeed, closed!

PR #1482, opened 2026-06-29
Amendment urls

tillprochaska:
This PR extends the ODPDocumentScraper to fetch amendment URLs. Most votes will only be about a single amendment, but there are cases where a vote may be about multiple amendments (when two or more semantically identical amendments are put to the vote together). For this reason, every vote can have a list of amendment URLs, rather than just a single amendment URL.

The amendments are linked in two places in the frontend:

* In the "More information" section of an amendment vote
* On the amendments page

I’ve added a note to all links (either as a description text below the link, or, in case of the amendments page, as a confirmation dialog after clicking the link) that highlights that the linked documents may contain multiple amendments.

You can run the following commands to update all votes with amendment URLs: (Update: I need to double-check this, but the VOTList pipeline below should work as expected.)

```
htv temp odp-documents
htv aggregate votes
```

Or run the VOTListPipeline for a single day:

```
htv pipeline vot-list --term=10 --date=2026-07-08
```

The last two commits contain a refactoring that is somewhat related to this PR, but I could also split it into a separate PR if that prevents conflicts with other PRs.


backend/howtheyvote/api/votes_api.py
```diff
@@ -834,6 +834,17 @@ def _format_related(votes: Iterable[Vote]) -> list[RelatedVoteDict]:
                 "amendment_subject": vote.amendment_subject,
                 "amendment_number": vote.amendment_number,
                 "amendment_authors": authors,
+                "amendment_urls": (
+                    [
+                        {
+                            "amendment_number": amendment_url.amendment_number,
```
linusha: I find it a bit strange that a vote potentially has a single amendment number (see line 835), but in here we could have multiple amendments numbers.
My understanding is that "equivalent" amendments will be voted on as one, and here we preserve the original amendment numbers - in such a case, what would be the number in line 835? 🤔 This is not critical for merging, just for me to understand it better...

backend/howtheyvote/scrapers/votes.py
```diff
@@ -719,6 +729,50 @@ def _eurovoc_concepts(self, doc: Any) -> tuple[set[str], set[str]]:
...
+        amendment_urls: list[AmendmentURL] = []
+        amendment_lists = doc["data"][0].get("inverse_foresees_change_of", [])
+
+        for amendment_number in amendment_numbers:
+            for amendment_list in amendment_lists:
+                start = amendment_list["itemNumberBegin"]
+                end = amendment_list["itemNumberEnd"]
+
+                # TODO Check if intervals are inclusive
+                if start > amendment_number or end < amendment_number:
```
linusha, line 753: This sounds like it still makes sense?

linusha (approved):

linusha:
> You can run the following commands to update all votes with amendment URLs: (Update: I need to double-check this, but the VOTList pipeline below should work as expected.)

I started this and aborted after a few minutes (just because it takes a long time to run), but I see no reason why this shouldn't work? :thinking: 

I ran the VOT pipeline for the last day of votes I have in my local DB and can confirm that that works.

tillprochaska:
> I started this and aborted after a few minutes (just because it takes a long time to run), but I see no reason why this shouldn't work? 🤔

I think it was a reminder to do a final test run to ensure that all edge cases are now properly handled. I think I did that and just forgot to update the PR description, but I will test one more time tomorrow.

backend/howtheyvote/scrapers/votes.py
tillprochaska: Ah, yes, I need to remove that comment, the intervals are indeed inclusive!

For example, the first item in

https://data.europarl.europa.eu/api/v2/plenary-documents/A-10-2026-0106?format=application/ld+json

has `itemNumberBegin=1` and `itemNumberEnd=6`, and the [corresponding PDF](https://redmapl3.europarl.europa.eu/RedmapFront/media/reds_iPlRp_Amd/A-10-2026-0106-AM-001-006/A-10-2026-0106-AM-001-006_en.pdf) contains exactly the six first amendments.

tillprochaska:
Did another full test run which completed successfully, so I’m going to merge and deploy this

PR #1493, opened 2026-07-06
Handle cases where we're granted access without solving a JS challenge

tillprochaska:
Not entirely sure when this happens, but I’ve observed this a few times today. Without this fix, we’d be waiting for a redirect (which never happens, because there was no JS challenge).

### DIFF
--- a/backend/howtheyvote/waf.py
+++ b/backend/howtheyvote/waf.py
@@ -1,4 +1,5 @@
 import ssl
+import time
 from typing import Any, cast
 
 import requests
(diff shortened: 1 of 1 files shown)


backend/howtheyvote/waf.py
```diff
@@ -167,3 +192,9 @@ def solve_aws_waf_challenge(url: str) -> str:
         },
     )
     raise WAFTokenError("Failed to obtain an AWS WAF token.")
+
+
+def _extract_token_from_cookies(cookies: list[dict[Any, Any]]) -> str | None:
```
linusha: This is unused right now? Did you commit half of a refactoring? I can see the code above 🤔

linusha (approved):

tillprochaska:
I opened this PR too early. I’ve updated it with an entirely new approach. This is also what has been deployed to production since beginning of the week.

linusha (approved):

PR #1497, opened 2026-07-08
Fix 10 second delay when restarting frontend container

tillprochaska:
When Docker restarts a container, it sends a SIGTERM signal. The root cause was probably related to the default init process in the container not properly forwarding this signal to the Node process. So instead of shutting down the Node process gracefully, Docker would wait for 10 seconds until timeouting and force-restarting the container. Setting `init: true` will include a proper `init` in the container that handles signals correctly.

### DIFF
--- a/frontend/scripts/watch
+++ b/frontend/scripts/watch
@@ -1,6 +1,5 @@
 #!/bin/bash
 
-trap exit SIGINT SIGTERM
 export ENTR_INOTIFY_WORKAROUND=1
 
 while true; do
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,7 @@ volumes:
 services:
   frontend:
     image: "ghcr.io/howtheyvote/howtheyvote-frontend:${HTV_TAG:-main}"
+    init: true
     depends_on:
       - backend
     expose:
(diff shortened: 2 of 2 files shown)


frontend/scripts/watch
```diff
@@ -1,6 +1,5 @@
 #!/bin/bash
 
-trap exit SIGINT SIGTERM
```
linusha, line 3: I am a bit confused as to why we don't need this anymore? 🤔 Or rather why we included it in the first place, when the container gets stopped (properly or by force) this script will be exited by definition, no?

linusha (approved):

frontend/scripts/watch
tillprochaska: Ah, yes, that’s a good question! This should in theory make the bash script exit when receiving one of the signals. The problem is that in the while loop, we’re starting another bash child process (the `-s` flag to `entr` means that the command will be evaluated using a shell). So I think the process hierarchy would look something like this:

1. shell (default init of the Docker image)
2. bash (the `watch` script)
3. `entr`
4. bash (started by `entr`)
5. node

`trap exit` only makes the first bash child process exit, but none of the other child processes. So the approach was the right direction, but it was never fully working. One could probably adjust this to make all process in the process group (including indirect children) exit, but using a proper init process (enabled by `init: true` in the Compose file) does handle all of this automatically.

PR #1499, opened 2026-07-09
Search improvements

tillprochaska:
This implements multiple small search improvements:

* Normalize special characters (e.g., "gregorova" should match "gregorová")
* Query spelling corrections
  * If there is a term that is similar to one of the query terms, display a suggestion (e.g., when searching for "defense", we ask "Did you mean defence?").
  * Because this is based only on string similarity, it sometimes results in weird suggestions. For now, the suggestion is only displayed if there are no more than 3 results for the original query.
  * If there is a term that is similar and there are no results for the original query, we directly apply the suggested alternative and show a notice (e.g., "Showing results for mercosur. There were no results for mercosour.").
  * I’ve also added logging for spelling corrections, so we can get an idea of what is suggested and fine tune based on that.
* Synonyms
  * I’ve implemented support for single-word synonyms (e.g., searching for "corona" finds votes containing the term "covid").
  * Synonyms can be configured globally via an environment variable. The format is `term1:synonym1|term2:synonym2`. This should be easy to output from Ansible.
* I’ve added a new feedback form for search results that is displayed at the end of the search results lists and when there are no results.
* I’ve added CLI commands to manually add search keywords to individual votes. This can be useful if a global synonym isn’t appropriate:
  * `htv search keywords add 123456 "chat control"`
  * `htv search keywords remove 123456 "chat control"`
  * `htv search keywords list 123456`

***

Todos (tests):

- [x] Normalize diacritics (gregorová -> gregorova)
- [x] Do not suggest stop words as correction (cat -> at)
- [x] Do not apply correction automatically if there are results for original query
- [x] Apply correction automatically if there are no results for original query
- [x] Only show suggestion if there are few results for the original query

***

Todos (post deployment):

- [ ] Adjust Ansible playbook to add new `HTV_SEARCH_SYNONYMS` env variable
- [ ] Reaggregate votes


linusha:
@tillprochaska I still have budget and saw that you asked for Copilot, which is why I ran it.

linusha:
@tillprochaska I think I would change the wording in the feedback form at the end of the vote results and have the second option be something like "No, I'm looking for votes on a broader topic" or something along these lines? Just having the negation makes it harder to understand, imo?

image

linusha:
**Update:** Nevermind, I just found that I need to run `htv system reload-synonyms` and not it works :tada: 

> Synonyms can be configured globally via an environment variable. The format is term1:synonym1|term2:synonym2. This should be easy to output from Ansible.

~~I can't get this to work :/~~
```
9dfc3f109843:/howtheyvote/backend# echo $HTV_SEARCH_SYNONYMS
corona:covid|chat control:eprivacy directive
```
~~but searching for "corona" yields results for corina. I reindexed after setting the env var. `htv search synonyms` also gives empty output.~~

linusha (approved):

linusha:
@tillprochaska one more thought. From the TODO above:

>  Adjust Ansible playbook to add new HTV_SEARCH_SYNONYMS env variable

This bags the question where the content lives with which this is populated, no? Would you put it into the playbook directly? Would be fine with me for now, if we actually end up having many, a separate text file in here that then gets inserted in the playbook might be more tidy? But that should no slow us down for now...

tillprochaska:
> I think I would change the wording in the feedback form at the end of the vote results and have the second option be something like "No, I'm looking for votes on a broader topic" or something along these lines? Just having the negation makes it harder to understand, imo?

Yes, I think that’s better wording. I’ve updated the form!

> This bags the question where the content lives with which this is populated, no? Would you put it into the playbook directly? Would be fine with me for now, if we actually end up having many, a separate text file in here that then gets inserted in the playbook might be more tidy?

My plan was to keep the list of synonyms in https://github.com/HowTheyVote/infra/blob/main/ansible/host_vars/htv-prod/vars.yml and not in this repo. The main motivation would be to make it very easy to update (without first making changes to this repo, rebuilding Docker images, etc.). What do you think?

linusha:
> The main motivation would be to make it very easy to update (without first making changes to this repo, rebuilding Docker images, etc.). What do you think?

That makes a lot of sense.
The one potential downside that comes to mind is that the infra repo is not public. But I think the pragmatic reasons are stronger - let's just put it in there @tillprochaska, I agree.

PR #1506, opened 2026-07-10
Caddy rate limiting

linusha:
Should be straight-forward for the most part, things to look out for:

1. I am unsure whether the adaption to the docker workflow is enough?
2. I placed the rate-limiting outside of the main config and only in the SITE_ADDRESS block in order to easily exempt our internal traffic. 
3. I have no hard opinion on the value of the actual rate-limiting. I began with 1/1s, that results in problems when opening /developers though, I think because there a redirect from /api/ to api/ happens or something similar. If you'd prefer a drastically lower limit, I am absolutely open to take an actual look at it, this workaround was good enough for me.
4. I also added a message mentioning the bulk download in the 429 response.

I did the following tests:

1. Hammering an internal route:

```
for i in {1..10}; do                                    
 wget localhost:5000/api/members/256810 --no-check-certificate
done
```
=> results in 10x 200er

2. Hammering the API via the SITE_ADDRESS:

```
or i in {1..10}; do
 wget localhost/api/members/256810 --no-check-certificate -O -
done
```

=> 5x 200er, afterwards 5 rate limited accesses, as expected.

3. Using the site under localhost (SITE_ADDRESS) worked, I opened up a few votes, /developers, did a search and opened an MEP profile.


tillprochaska:
Nice, I will take a look! Regarding the concrete value of the rate limits: We could also combine multiple windows. For example, at most X requests per second and at most Y requests per minute. This would allow shorter bursts which shouldn’t be a problem.

caddy/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM caddy:2.11.4-builder-alpine AS builder
+
+RUN xcaddy build \
+    --with github.com/mholt/caddy-ratelimit
+
+FROM caddy:2.11.4-alpine
+
+COPY --from=builder /usr/bin/caddy /usr/bin/caddy
```
tillprochaska: Trailing newline for consistency with other Dockerfiles

```suggestion
COPY --from=builder /usr/bin/caddy /usr/bin/caddy

```

caddy/Caddyfile
```diff
@@ -95,6 +95,23 @@
...
+                path /api/*
+                path /api
+            }
+        }
+    }
+
+    handle_errors 429 {
+        header Content-Type "application/json"
+        respond `{"error": "rate_limited", "message": "For a bulk download of our data, please visit https://github.com/howtheyVote/data"}` 429
+    }
```
tillprochaska, line 123: If we extend this to the frontend as well, it would be nice to return a JSON response for requests to `/api` while returning an HTML response for other requests. There might be situations where normal users run into the rate limit. For example, I sometimes open a lot of links in a new tab, just to look at them one after another. Or users in a corporate network might share the same IP address. For these cases, we should display a human-friendly error message and possibly include information on how to get in contact if a user thinks their usage was normal. Ideally, the HTML response would look similar to our other error pages.

caddy/Caddyfile
```diff
@@ -95,6 +95,23 @@
...
+            match {
+                path /api/*
+                path /api
+            }
+        }
+    }
+
+    handle_errors 429 {
+        header Content-Type "application/json"
+        respond `{"error": "rate_limited", "message": "For a bulk download of our data, please visit https://github.com/howtheyVote/data"}` 429
```
tillprochaska: Maybe include information about the rate limits in the error message, so users know what rates are acceptable. (Could also be lower than the enforced limit.)

caddy/Caddyfile
```diff
@@ -95,6 +95,23 @@
...
 
 {$CADDY_SITE_ADDRESS} {
+    rate_limit {
+        zone dynamic {
+            key {remote_host}
+            window 10s
+            events 5
+            match {
+                path /api/*
+                path /api
```
tillprochaska, line 115: We can start with this, but according to our metrics, a lot of the bot traffic is just crawling the frontend pages not the API directly. So we’d probably have to enforce some limit on the frontend as well.

If we do this, we should also ensure that we exempt requests to static resources (CSS/JS + files). When loading a single vote page, this can easily trigger a lot of requests to static resources (e.g., MEP photos).

tillprochaska (commented):
Just a few quick thoughts, haven’t looked at it in detail yet.

caddy/Caddyfile
linusha: Yes, I agree! I'll add after we decided on a limit, ok?

linusha:
@tillprochaska I tested from the frontend container as well with caddy:80 and I think everything works :)

.github/workflows/docker.yml
```diff
@@ -19,7 +19,9 @@ jobs:
         image:
           - frontend
           - backend
-
+          - caddy
+    # Only build caddy image when we really need to
+    if: matrix.image != 'caddy' || github.event_name == 'workflow_dispatch'
```
tillprochaska: Nice!

tillprochaska:
```suggestion

    # Only build caddy image when we really need to
    if: matrix.image != 'caddy' || github.event_name == 'workflow_dispatch'

```

caddy/Caddyfile
```diff
@@ -95,6 +95,23 @@
 }
 
 {$CADDY_SITE_ADDRESS} {
+    rate_limit {
+        zone dynamic {
```
tillprochaska:
```suggestion
        zone api {
```

caddy/Dockerfile
tillprochaska, line 1: Can you update the Dependabot configuration so it checks for updates to the Caddy base image? (You can probably copy-paste the configuration for the frontend or backend Dockerfiles.)

tillprochaska (approved):
I think merging the rate limits for the APIs is a good first step. I’d suggest combining two windows, something like this:

* 20 requests per second
* 120 requests per minute

This means if you’re just hacking together a quick script to fetch a small amount of data you’re not impacted. But continuously sending requests at a high rate will get you blocked. If we noticed that one of the limits is too generous, we can still decrease them later?

I think it should be possible to implement this by defining two separate zones with different names? (If it’s more complex than that, let’s ignore it for now.)

caddy/Caddyfile
tillprochaska: Oh, and I think JSON response like this is a bit more common (the structure, not the indentation).

```json
{
  "error": {
    "code": "TOO_MANY_REQUESTS",
    "message": "Rate limit exceeded. Please send no more than ..."
  }
}
```

caddy/Caddyfile
tillprochaska: An idea on how to handle frontend rate limits:

1. Enforce a normal rate limit
2. If a user hits the rate limit, show an error page.
3. The error page has a button. After clicking on the button a cookie is set.
4. If the cookie is set, the rate limit is bypassed.

This is of course super easy to circumvent. But it might be enough to exclude the dumb, generic scrapers that cause excessive load.

caddy/Caddyfile
linusha: Oh yeah, I somehow thought this was a keyword....

linusha:
@tillprochaska this should be ready for merge, I left it open because I did not want to deploy this without being able to check that everything is fine (or intervening if not). If you're going to deploy something feel free to include it, otherwise I'll do it soon.

linusha:
@tillprochaska okay, now it should be ready for real - the condition inside of the docker workflow previously did not work as expected. I trusted Claude there and to confirm I simply looked at the green checks at the bottom of the PR - that was wrong though, the workflow became invalid as of that commit, see [this run](https://github.com/HowTheyVote/howtheyvote/actions/runs/29083958792). It's not possible to skip entire jobs inside of a matrix without duplicating the if statement throughout all steps. Imo, the solution I implemented now is cleaner than duplicating the condition many times over - and the green runs are all back also in the PR :)

PR #1512, opened 2026-07-13
Clarify rejection

### LINKED ISSUE #1510: Banner for rejection votes
Display a banner indicating that a vote is a motion to reject and that this needs to be taken into account when interpreting voting behavior.

### LINKED ISSUE #1511: Don't show snippet for rejection votes

tillprochaska:
There are three cases:

* We have information about the procedure stage:
  * 1st reading: If adopted, President asks Commission to withdraw
  * 2nd reading: If adopted, the legislative procedure is closed.
* We don’t have information about the procedure stage: It could be either 1st or 2nd reading, so we display a more generic message.

To test this locally:

* Run the main votes analyzer: `htv temp main-votes`
* Aggregate votes: `htv aggregate votes`

Example votes:

* No procedure stage: http://localhost:8000/votes/179804
* 1st reading: http://localhost:8000/votes/180983
* 2nd reading: http://localhost:8000/votes/195775


frontend/src/pages/VotePage.tsx
```diff
@@ -90,6 +95,40 @@ export const VotePage: Page<Vote> = ({ data }) => {
...
+        {isRejection && (
+          <>
+            {data.procedure?.type === "COD" &&
+              data.procedure?.stage === "OLP_FIRST_READING" && (
+                <Callout>
+                  <p>
+                    This vote is about a proposal to reject a draft legislation.
+                    If the proposal to reject secures a simple majority of votes
+                    cast, the President of the European Parliament will ask the
+                    Commission to withdraw the draft.
```
tillprochaska: In theory, a legislative proposal isn’t guaranteed to have been proposed by the Commission:

> The Commission holds the ‘right of initiative’, i.e. the prerogative to propose
legislation at Union level (Article 17 TEU). In a limited number of specific cases, the Treaties provide that a proposal for a legislative act may also be submitted on the initiative of a group of Member States, on a recommendation by the European Central Bank, or at the request of the Court of Justice (Article 294(15) TFEU).

https://www.europarl.europa.eu/cmsdata/304031/OLP%20Handbook_September%202025_EN.pdf#page=27

I think it would be nice to keep the text as simple as possible (rather than using more generic wording such as "The President will ask the originating instiution to withdraw…".

However, it would be good to check how common the exceptions quoted above are and whether there are any such cases in our database.

frontend/src/pages/VotePage.tsx
```diff
@@ -90,6 +96,40 @@ export const VotePage: Page<Vote> = ({ data }) => {
...
+                </Callout>
+              )}
+            {data.procedure?.type === "COD" &&
+              data.procedure?.stage === "OLP_SECOND_READING" && (
+                <Callout>
+                  <p>
+                    This vote is about a proposal to reject the Council’s
+                    position. If the proposal to reject secures an absolute
+                    majority of Parliament’s component members, the legislative
+                    procedure ends.
```
tillprochaska: While it’s the official wording, I think "absolute majority of Parliament’s component members" is a bit clunky. Maybe we could simplify that?

> If more than half of all MEPs vote in favour of the proposal to reject…

frontend/src/pages/VotePage.tsx
tillprochaska: Add `showSnippet={!isRejection}` property to VoteHeader

frontend/src/pages/VotePage.tsx
```diff
@@ -64,6 +64,12 @@ export const VotePage: Page<Vote> = ({ data }) => {
   // This will not work in all cases, but might be good enough in combination with the amendment condition above.
   const main_vote = data.related.find((related_vote) => related_vote.is_main);
 
+  const isRejection =
+    data.is_main &&
+    data.procedure?.type === "COD" &&
+    (data.description === "Rejet" ||
```
linusha: I don't think this condition is sufficient? We have quite a few instances like [this one](https://howtheyvote.eu/votes/164096), where I'd think we need a `startsWith("rejet")` as well?

frontend/src/pages/VotePage.tsx
linusha: I think "will ask for the draft to be withdrawn" dodges this problem and does not make the sentence more complicated?

tillprochaska: Oh yes, that is obviously better!

frontend/src/pages/VotePage.tsx
tillprochaska: Oh right, I thought the description would be only the part before the first dash, but I confused this with the MainVoteAnalyzer.

frontend/src/pages/VotePage.tsx
tillprochaska: @linusha Do you remember which wording we ended up with here?

I think we discussed that this alone might not be clear enough?

> This vote is about a proposal to reject the Council’s position. If more than half of all MEPs vote in favour of the proposal, the legislative procedure ends.

tillprochaska: I feel like trying to make this explicit makes the sentence more confusing 🙃 

> This vote is about a proposal to reject the Council’s position. If an absolute majority of all MEPs (not just a majority of those present) vote in favour of the proposal, the legislative procedure ends.

linusha: I think we thought about "all elected", and then made jokes about :es:.

But on a constructive note: I tend to just go with the absolute number? Could only change every 5 years and even with elections, a change in the size of Parliament is the outlier and not the norm?

frontend/src/pages/VotePage.tsx
tillprochaska: Fixed

frontend/src/pages/VotePage.tsx
tillprochaska: Good point, I’ve updated the text. I wasn’t entirely sure whether it would be 719/2+1 or 720/2+1. According to [this article](https://www.patrick-breyer.de/eu-parlament-winkt-chatkontrolle-1-0-durch-breyer-wahrer-verlierer-sind-unsere-kinder/) it’s 720/2+1.

tillprochaska: I guess "component members" refers to the seats allocated by the Council not whether a seat is actually filled, but I haven’t found an explicit formal basis for this.

frontend/src/pages/VotePage.tsx
linusha: > However, it would be good to check how common the exceptions quoted above are and whether there are any such cases in our database.

I looked into this, @tillprochaska - most importantly, these exceptions have super limited scopes.

For the ECB and the CJEU it seems that the exception is about changing their own statutes. We have [a vote](https://howtheyvote.eu/votes/164466) on that for the CJEU! Last time I found for the ECB was from 2017.

For the member states [Source](https://academic.oup.com/oxford-law-pro/book/58863/chapter-abstract/492293499?redirectedFrom=fulltext):

> Moreover, Article 76 TFEU applies only to measures in the fields of judicial cooperation in criminal matters and police cooperation.

[Here](https://eur-lex.europa.eu/eli/dir/2014/41/oj/eng) is an example. 

So honestly, saying that the Commission has the (sole) right to initiate laws is really only wrong on a technical level, imo. But let's just keep the generic sentence, one thing less to worry about.

Super interesting side-quest!

frontend/src/pages/VotePage.tsx
linusha: I also was not able to find a source for this, but I also don't think it matters in this case, as 719/2+21=360.5 which would round up to 361 for a majority anyways.

frontend/src/pages/VotePage.tsx
linusha: Took the liberty of quickly adding this to this PR. You can check with [snippet](http://localhost:8000/votes/182886) and [without](http://localhost:8000/votes/182885).

linusha (approved):
I smuggled something onto your branch, hope that's ok.

Feel free to 🚢!

PR #1514, opened 2026-07-15
Return 404 if vote ID is invalid

tillprochaska:


frontend/src/pages/VotePage.tsx
```diff
@@ -25,6 +26,14 @@ import type { Loader, Page, Request } from "../lib/server";
 const log = getLogger();
 
 export const loader: Loader<Vote> = async (request: Request) => {
+  // For some reason, we get a bunch of requests like this: https://howtheyvote.eu/votes/123456.csv.
+  // The vote ID is valid, so the subsequent API request would succeeded, but return CSV instead of
```
linusha:
```suggestion
  // The vote ID is valid, so the subsequent API request would succeed, but return CSV instead of
```

frontend/src/pages/VotePage.tsx
```diff
@@ -25,6 +26,14 @@ import type { Loader, Page, Request } from "../lib/server";
...
+  // a JSON response. This results in a lot of errors in Sentry. (The proper solution would be to
```
linusha, line 31: > The vote ID is valid, so the subsequent API request would succeeded, but return CSV instead of a JSON response.

I am a bit confused by this sentence? It does not return a CSV that's kind of the point, no? People might think that it does?

linusha (approved):

frontend/src/pages/VotePage.tsx
tillprochaska: The frontend doesn’t, but a request to https://howtheyvote.eu/votes/194553.csv would trigger an API request to https://howtheyvote.eu/api/votes/194553.csv, and the API returns CSV. So the frontend receives CSV, but expects JSON.

linusha: Oh of course, yes that makes sense!

PR #1517, opened 2026-07-15
Fix data workflow docker

tillprochaska:


linusha:
@tillprochaska do you have an idea why this broke? It ran [a week ago](https://github.com/HowTheyVote/howtheyvote/actions/runs/28945741948) :O

**edit:** I assume it's due to the restructuring of the docker files in the alpine upgrade PR! Only saw that after writing here...

tillprochaska:
Yes, exactly, I forgot to handle this properly and noticed only after merging the changes.

PR #1526, opened 2026-07-17
Simplify docker workflow after adding build step to separate workflows

### LINKED ISSUE #1527: Use path filters for pull_request triggers in frontend/backend workflows
These workflows are now a bit more expensive, so it might make sense to only run them when necessary. For example, we only need to run the frontend CI when there are changes in the frontend directory.

linusha:
Removing the  pull_request trigger as mentioned in #1050 and also the workaround to conditionally skip the caddy build as we don't need that when the pull_request trigger is gone.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,32 +4,21 @@ on:
   workflow_dispatch: {}
   push:
     branches: ["main"]
-  pull_request: {}
 
 permissions:
   contents: read
   packages: write
 
 jobs:
-  prepare:
-    runs-on: ubuntu-latest
-    outputs:
-      images: ${{ steps.set.outputs.images }}
-    steps:
-      - id: set
-        run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -4,32 +4,21 @@ on:
   workflow_dispatch: {}
   push:
     branches: ["main"]
```
linusha: We have this because we don't enforce branches to be rebased before merging, right? And thus the PR checks don't guarantee that main will build after merging?

tillprochaska: Yes, I think so. I think we could probably remove it as well, because the frontend/backend workflows also run for the main branch, and they now include building the images.

tillprochaska (approved):

tillprochaska:
@linusha I just thought of this which is related. If you feel like it you could make that change in this PR as well. And if not, there’s a ticket for it so we don’t forget :)

https://github.com/HowTheyVote/howtheyvote/issues/1527

PR #1528, opened 2026-07-17
Generate share-pictures for date when specified

linusha:
The alternative was always editing the command code from hand which became a bit tedious...


tillprochaska (approved):
(Looks good, but haven’t tested myself.)

PR #1530, opened 2026-07-17
Intercept HTTP requests in worker tests

tillprochaska:
I noticed that the `rcv_list_pipeline` test was very slow. It turned out that the VOTListPipeline wasn’t mocked, so the worker would execute it in between the RCVListPipeline runs against real data. This fixes that by intercepting all HTTP requests.

### DIFF
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -64,7 +64,7 @@ class DummyRegistry(FirstMatchRegistry):
         return response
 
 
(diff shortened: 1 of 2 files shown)


tillprochaska:
Wait, maybe there should be a more generic solution that intercepts all HTTP requests by default and does not require explicitly using the responses fixture

linusha (approved):
This is pretty cool! Great find!

PR #1531, opened 2026-07-17
Remove unnecessary packages

tillprochaska:
I think these were necessary some time ago because some of the Python packages needed to be built from source (e.g. cryptography as a transitive dependency). It seems that all of these packages are now available as pre-built binaries, so the build-time dependencies aren’t required anymore.

This reduces the backend image size from 1.9 GB to 800 MB.

I tested the following:

* Stop the backend container: `docker compose down backend`
* Rebuild backend image: `docker compose build --no-cache backend`
* Remove the virtualenv directory (`.venv`)
* Reinstall Python packages: `docker compose run --rm backend poetry install`
* Start the backend container: `docker compose up -d --force-recreate backend`
* Run the tests: `docker compose exec backend make test`
* Run a pipeline: `docker compose exec backend htv pipeline rcv-list --term=10 --date=2026-05-21`

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -45,23 +45,11 @@ FROM python:3.14-alpine3.24
 
 ARG PYTHON_VERSION
 
-RUN apk add --no-cache \
-    build-base \
-    libffi-dev \
-    # TODO: Switch back to LibreSSL once conflicts on Alpine 3.17 are resolved.
-    openssl-dev \
-    libxml2-dev \
-    libxslt-dev \
-    python3-dev \
-    bash \
-    less \
-    make \
-    cargo \
-    sqlite \
-    tmux
+RUN apk add --no-cache libstdc++ bash less make sqlite tmux
 
 # Copy Xapian and the Python bindings from the builder stage
 COPY --from=builder /usr/local/lib/libxapian.so* /usr/local/lib/
+COPY --from=builder /usr/lib/libpython${PYTHON_VERSION}.so.1.0 /usr/lib/
 COPY --from=builder /usr/local/lib/python${PYTHON_VERSION}/site-packages/xapian /usr/local/lib/python${PYTHON_VERSION}/site-packages/xapian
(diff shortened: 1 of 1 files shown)


backend/Dockerfile
```diff
@@ -45,23 +45,11 @@ FROM python:3.14-alpine3.24
...
-    less \
-    make \
-    cargo \
-    sqlite \
-    tmux
+RUN apk add --no-cache libstdc++ bash less make sqlite tmux
 
 # Copy Xapian and the Python bindings from the builder stage
 COPY --from=builder /usr/local/lib/libxapian.so* /usr/local/lib/
+COPY --from=builder /usr/lib/libpython${PYTHON_VERSION}.so.1.0 /usr/lib/
```
linusha: Can you explain to me why this was added here?

And I assume `libstdc++` above is also due to some Xapian-building-stuff?

tillprochaska: Ah, good question. Yes, Xapian needs the C++ standard library. It was previously installed as a transitive dependency by `build-base` (and possibly others).

The Xapian Python bindings are linked against `libpython`. That was previously installed by `python3-dev`. After removing `python3-dev`, importing the Xapian package failed due to it not finding `libpython`, so I copied it over from the builder image.

But I just realized that’s not necessary. In the builder image, it’s also installed via the system package manager (via `python3-dev`), so it’s saved in `/usr/lib`. And probably the linker linked the Python bindings to that version, though I’m not entirely sure about that. But in the runtime image, only the version from the base image at `/usr/local/lib` is present. It should be enough to tell the linker to use `/usr/local/lib` when linking the bindings in the builder stage.

tillprochaska: Pushed a commit to fix that

tillprochaska: Reduced the image size by another 5 MB :D

tillprochaska:
A nice side-effect is that this also speeds up CI runs because we need to download/upload less data from/to the GitHub Actions cache.

linusha (approved):
Can confirm this works on my machine, too!

PR #1535, opened 2026-07-30
WAF for Sessions and Members Scrapers

linusha:
This should solve the failure we saw on Monday. For the member scraper, I needed to implement logic to retrieve new tokens, as a single one would yield to the challenge being presented again after a while.

My [testrun] is still going, afterwards we might be able to gauge how long a single token is good for.

We should probably aim at moving these two pipelines towards the ODP even more, but for the moment this should be an acceptable workaround I'd think...

### DIFF
--- a/backend/howtheyvote/pipelines/members.py
+++ b/backend/howtheyvote/pipelines/members.py
@@ -21,6 +21,7 @@ from ..scrapers import (
 )
 from ..sharepics import generate_member_sharepic
 from ..store import Aggregator, BulkWriter, index_records, map_member
+from ..waf import solve_ep_aws_waf_challenge
 from .common import BasePipeline
 
 log = get_logger(__name__)
@@ -33,6 +34,7 @@ class MembersPipeline(BasePipeline):
         self._member_ids: set[str] = set()
 
     def _run(self) -> None:
+        self._ep_aws_waf_token = solve_ep_aws_waf_challenge()
         self._scrape_members()
         self._scrape_member_groups()
         self._scrape_member_infos()
@@ -44,7 +46,7 @@ class MembersPipeline(BasePipeline):
         log.info("Scraping members", term=self.term)
 
         writer = BulkWriter()
-        scraper = MembersScraper(term=self.term)
+        scraper = MembersScraper(term=self.term, aws_waf_token=self._ep_aws_waf_token)
         writer.add(scraper.run())
         writer.flush()
 
@@ -57,7 +59,9 @@ class MembersPipeline(BasePipeline):
             log.info("Scraping member groups", term=self.term, member_id=member.id)
 
             try:
-                scraper = MemberGroupsScraper(web_id=member.id, term=self.term)
+                scraper = MemberGroupsScraper(
+                    web_id=member.id, term=self.term, aws_waf_token=self._ep_aws_waf_token
+                )
                 writer.add(scraper.run())
             except ScrapingError as err:
                 log.exception(
@@ -76,7 +80,9 @@ class MembersPipeline(BasePipeline):
             log.info("Scraping member info", term=self.term, member_id=member.id)
 
             try:
-                scraper = MemberInfoScraper(web_id=member.id)
+                scraper = MemberInfoScraper(
+                    web_id=member.id, aws_waf_token=self._ep_aws_waf_token
+                )
                 writer.add(scraper.run())
             except ScrapingError as err:
                 log.exception(
--- a/backend/howtheyvote/pipelines/members.py
+++ b/backend/howtheyvote/pipelines/members.py
@@ -18,6 +18,7 @@ from ..scrapers import (
     MemberInfoScraper,
     MembersScraper,
     ScrapingError,
+    WAFChallengeError,
 )
 from ..sharepics import generate_member_sharepic
 from ..store import Aggregator, BulkWriter, index_records, map_member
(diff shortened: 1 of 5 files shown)


backend/howtheyvote/pipelines/members.py
```diff
@@ -54,18 +57,31 @@ def _scrape_member_groups(self) -> None:
...
-                sentry_sdk.capture_exception(err)
+            for attempt in range(1, 2):
+                try:
+                    scraper = MemberGroupsScraper(
+                        web_id=member.id, term=self.term, aws_waf_token=self._ep_aws_waf_token
+                    )
+                    writer.add(scraper.run())
+                    break
+                except WAFChallengeError as waf_err:
+                    if attempt == 2:
```
tillprochaska, line 68: Does it make sense to continue with the next member in this case? I’d expect if the token doesn’t work for the current member, then it also wouldn’t work for the next, so we could just fail the entire pipeline at this point?

backend/howtheyvote/pipelines/members.py
```diff
@@ -54,18 +57,31 @@ def _scrape_member_groups(self) -> None:
...
-                scraper = MemberGroupsScraper(web_id=member.id, term=self.term)
-                writer.add(scraper.run())
-            except ScrapingError as err:
-                log.exception(
-                    "Failed scraping member groups",
-                    member_id=member.id,
-                    term=self.term,
-                )
-                sentry_sdk.capture_exception(err)
+            for attempt in range(1, 2):
```
tillprochaska, line 60: I think the end is exclusive, so currently, the loop iterates only once.


```suggestion
            for attempt in range(2):
```

tillprochaska: No strong preference, but separating the token retries from the scraping errors might make this a bit easier to follow, something like this. Then you also don’t have to break explicitly in case of a `ScrapingError`

```py
for member in self._members():
  try:
    for attempt in range(2):
      try:
        # Run scraper
      except WAFChallengeError:
        if attempt == 0:
          # Get new token
        else:
          # New token failed as well
  except ScrapingError:
    # handle as usual
    
```

backend/howtheyvote/pipelines/members.py
```diff
@@ -54,16 +57,38 @@ def _scrape_member_groups(self) -> None:
...
+                            time.sleep(config.TOKEN_RENEWAL_SLEEP)
+                            log.warning(
+                                "Obtaining new WAF token", member_id=member.id, attempt=attempt
+                            )
+                            self._ep_aws_waf_token = solve_ep_aws_waf_challenge()
+                            continue
+            except WAFChallengeError as waf_err:
+                log.exception("New WAF token did not solve challenge", member_id=member.id)
+                sentry_sdk.capture_exception(waf_err)
+                raise
```
linusha, line 89: I think I need to handle this outside of the loop explicitly once more, as `WAFChallengeError` is a subclass of `ScrapingError`.

linusha:
:rotating_light: I just deployed this once more and am currently running the Members pipeline. (Unfortunately, just setting the prod machine as exit node in tailscale did not work for me, but we can figure this out later).

linusha:
@tillprochaska this ran through, at least the scraping part. It failed when trying to download the pictures, but I unfortunately cannot access the error message anymore because I printed too many other lines too fast :/

However, having the members is the relevant part for the other things to work, pictures are not mission critical.

We should investigate that, but as far as I can see, this works. I am not so sure why I can't see the image problem - whatever it was - in sentry? :thinking:

linusha:
I looked into the image error - it's just a result of the image file that is written being empty because we are presented a waf challenge there, too.

```
linus@htv-prod:/usr/local/howtheyvote/storage/files/members$ file 101039.jpg
 101039.jpg: empty
linus@htv-prod:/usr/local/howtheyvote/storage/files/members$ curl -I https://www.europarl.europa.eu/mepphoto/101039.jpg
 HTTP/2 202 
 server: CloudFront 
 date: Mon, 14 Sep 2026 08:25:45 GMT
 content-length: 0 
 x-amzn-waf-action: challenge
 cache-control: no-store, max-age=0
 content-type: text/html; charset=UTF-8 
 access-control-allow-origin: *
 access-control-max-age: 86400
 access-control-allow-methods: OPTIONS,GET,POST
 access-control-expose-headers: x-amzn-waf-action
 x-cache: Error from cloudfront
 via: 1.1 55107fc1be09ed1afcf3154ed9bd93cc.cloudfront.net (CloudFront)
 x-amz-cf-pop: FRA56-P5
 x-amz-cf-id: BmBbWc0eBsMdgeMxC-uWnqqq6rwygoy1V8ImjMvUMgPHhESca1BIew==
```

backend/howtheyvote/pipelines/members.py
```diff
@@ -74,10 +99,33 @@ def _scrape_member_infos(self) -> None:
...
+                    try:
+                        scraper = MemberInfoScraper(
+                            web_id=member.id, aws_waf_token=self._ep_aws_waf_token
+                        )
+                        writer.add(scraper.run())
+                        break
+                    except WAFChallengeError:
+                        if attempt == 1:
+                            raise
+                        else:
```
tillprochaska: `else` is unnecessary, as the `if` block raises

backend/howtheyvote/pipelines/members.py
```diff
@@ -74,10 +99,33 @@ def _scrape_member_infos(self) -> None:
...
+                                "Waiting before trying to obtain new token.",
+                                member_id=member.id,
+                                attempt=attempt,
+                            )
+                            time.sleep(config.TOKEN_RENEWAL_SLEEP)
+                            log.warning(
+                                "Obtaining new WAF token", member_id=member.id, attempt=attempt
+                            )
+                            self._ep_aws_waf_token = solve_ep_aws_waf_challenge()
+                            continue
```
tillprochaska: `continue` is unnecessary as it’s the last statement in the loop body?

backend/howtheyvote/pipelines/members.py
tillprochaska: Just to confirm my understanding is correct: We need to handle it separately because otherwise the generic `ScrapingError` block would handle the exception and continue with the next MEP, but we want to fail the entire pipeline, right?

tillprochaska (approved):
Just two minor suggestions, otherwise looks good

tillprochaska:
> We should investigate that, but as far as I can see, this works. I am not so sure why I can't see the image problem - whatever it was - in sentry? 🤔

I can see one instance of it in Sentry from this morning. Did you fix something which caused it to not be visible in Sentry? Or is this still an open question?

backend/howtheyvote/pipelines/members.py
linusha: Yes, exactly.

linusha:
When the error occurred after the pipeline ran bc of the scheduling, it appeared in Sentry. 

When I ran it one-off from the CLI in prod yesterday, the error was the same but it was not recorded in Sentry.

PR #1551, opened 2026-09-03
Fix rounding in vote result charts

tillprochaska:
A user reported that we’re sometimes rounding the percentages incorrectly (see test cases for an example). The reason for that is that we were rounding twice, first to 3 decimal places, then again to 2.

### DIFF
--- a/frontend/src/components/VoteResultChart.tsx
+++ b/frontend/src/components/VoteResultChart.tsx
@@ -51,12 +51,12 @@ function Summary({ stats }: SummaryProps) {
 
   return (
     <p class="text--sm text--light" data-testid="vote-result-chart-summary">
-      For: <strong>{stats.FOR}</strong>
-      <span class="visually-hidden">({percentageFor} %)</span>. Against:{" "}
-      <strong>{stats.AGAINST}</strong>
-      <span class="visually-hidden">({percentageAgainst} %)</span>. Abstentions:{" "}
-      <strong>{stats.ABSTENTION}</strong>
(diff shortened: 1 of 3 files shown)


frontend/src/components/VoteResultChart.tsx
linusha: I tried this locally and think it should always work, but maybe I am not seeing something - can't this just go then and we use `percentage` in `--ratio` as well?

linusha, line 38:
```suggestion
      style={{ "--ratio": percentage }}
```

This is what I meant above.

linusha (approved):
If we can get rid of the other calculation great, but this works in any case :)

frontend/src/components/VoteResultChart.tsx
tillprochaska: `ratio` is a value between 0 1 with 3 decimal places, and `percentage` is a rounded integer value on a scale from 0 to 100. So doing this technically means that the `width` of all bars will sum up to 10,000%. In practice, flexbox shrinks all bars proportionally, so the bars do not overflow the container. Still, `ratio` is more precise than `percentage`, resulting in slight visual differences for small bars.

tillprochaska: But I’ve added a comment to clarify the difference

linusha: I was aware of this difference, I just tested plunging in less precise values into the example that previously didn't match 100 and thought that it would look the same for all values. 
Might also be that it looked slightly different and I didn't notice. If you say it makes a visual difference, we of course need to keep it - that was the bagged question.

PR #1553, opened 2026-09-14
Pass WAF token when downloading MEP photos and in PressPipeline

tillprochaska:
### `run_each_with_waf_token`

I’ve added a new helper `run_each_with_waf_token(iterable, func, current_waf_token, solve_waf_challenge, sleep)`. It calls `func` with every item in `iterable` and the WAF token. If this raises a `WAFChallengeException`, it waits for `sleep` seconds, then calls `solve_waf_challenge` to retrieve a fresh WAF token and retries the `func` call one time. 

This new helper is now used in all places that need to scrape data/download files in a loop (such as the `MembersPipeline` and `PressPipeline`).

### `WAFChallengeError`

`WAFChallengeError` (the exception that is raised when an HTTP request fails because we’re prompted to solve a WAF challenge) is not a subclass of `ScrapingError` anymore.

The original motivation for this change was that I wanted to use it outside of scrapers (in the `download_file` helper used to download the MEP photos), but it actually helps simplify the code in some places: When we handle `ScrapingErrors` without failing the entire pipeline, we don’t need to explicitly re-raise `WAFChallengeError` anymore.

One downside is that it might be a little confusing that `WAFChallengeError` is not a subclass of `ScrapingError`, or rather that scrapers may raise exceptions other than `ScrapingError`?

### `download_file`

`download_file` now also accepts `waf_token` as an argument. It will now use the same custom requests `Session` object (which mimics a real browser and automatically injects the WAF token) to download MEP photos.

I’ve also moved the check for the AWS WAF header that indicates a challenge into a [`Session` hook](https://requests.readthedocs.io/en/latest/user/advanced/#event-hooks). Otherwise, it would’ve been necessary to duplicate it for `get_url` (in the `scrapers.common` module) and `download_file` (in the `files` module).

## Testing

I have tested the following locally:

- [x] Running press pipeline with `htv pipeline press --rss --date=2026-09-15`
- [x] Running members pipeline with `htv pipeline members --term=10`


backend/howtheyvote/pipelines/members.py
```diff
@@ -56,78 +59,41 @@ def _scrape_members(self) -> None:
...
-                log.exception("New WAF token did not solve challenge", member_id=member.id)
-                sentry_sdk.capture_exception(waf_err)
-                raise
+                scraper = MemberGroupsScraper(
+                    web_id=member.id,
+                    term=self.term,
+                    aws_waf_token=waf_token,
+                )
+                writer.add(scraper.run())
             except ScrapingError as err:
```
tillprochaska: Need to re-raise WAFChallengeError

backend/howtheyvote/pipelines/members.py
```diff
@@ -56,78 +59,41 @@ def _scrape_members(self) -> None:
...
-                            attempt=attempt,
-                        )
-                        self._ep_aws_waf_token = solve_ep_aws_waf_challenge()
-            except WAFChallengeError as waf_err:
-                log.exception("New WAF token did not solve challenge", member_id=member.id)
-                sentry_sdk.capture_exception(waf_err)
-                raise
+                scraper = MemberInfoScraper(web_id=member.id, aws_waf_token=waf_token)
+                writer.add(scraper.run())
             except ScrapingError as err:
```
tillprochaska: Need to re-raise WAFChallengeError

tillprochaska:
Still need to run the full MembersPipeline once for testing. But otherwise it’s ready.

backend/howtheyvote/waf.py
```diff
@@ -172,3 +190,25 @@ def solve_aws_waf_challenge(url: str, timeout: int = 60) -> str:
...
+    new WAF token and retry once."""
+    waf_token = current_waf_token
+
+    for item in iterable:
+        try:
+            func(item, waf_token)
+        except WAFChallengeError:
+            time.sleep(sleep)
+            waf_token = solve_waf_challenge()
+            func(item, waf_token)
```
linusha, line 212: The way its limited to one retry is because we call `func` from the expect block and the next raise will not be excepted, right?

backend/howtheyvote/waf.py
```diff
@@ -172,3 +190,25 @@ def solve_aws_waf_challenge(url: str, timeout: int = 60) -> str:
...
+
+    return waf_token
```
linusha, line 214: At first I was a bit confused about why this returns the token and why the output of e.g. scraping all member groups is the token again (which is just not what you'd expect semantically) but: This just allows us to reuse the token for the next steps in the pipeline, right? So e.g. we get a new token, the current iterator is processed and then we will first try to reuse this token for the next steps, before actively trying to solve another challenge without being prompted for it?

linusha (approved):
I did not run this, mainly my review is a "do I understand what is happening".

Also, I don't think its particularly confusing that scrapers now can raise exceptions other than `ScrapingException`. I get how this might be a bit unsatisfactory, but with the current state of affairs, I don't think we will forget about the waf problems and then be confused later 🙃

backend/howtheyvote/waf.py
tillprochaska: Correct!

backend/howtheyvote/waf.py
tillprochaska: Correct!

